PR #2, opened 2011-01-11
Turns out seedRandom doesn't work on unix platforms. This fixes that

DRMacIver:
Totally trivial to fix. Added a call to seedRandom with a numeric argument in the tests file so that it would get caught if it happened again.

### DIFF
--- a/lib-clay/random/platform/platform.unix.clay
+++ b/lib-clay/random/platform/platform.unix.clay
@@ -12,5 +12,5 @@ seedRandom(x:UInt) {
 
 seedRandomWithTime() {
     var t = unix.time(null(CLong));
-    seedRandom(Int(t));
+    seedRandom(UInt(t));
 }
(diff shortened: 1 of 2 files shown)


jckarter:
test/random.clay doesn't build with your patch (at least on OS X) because there are some type mismatches remaining. Could you fix them?

Now that I look at it, the implementation of random() in the random module is totally bogus and doesn't return uniformly distributed values. uniform_random(a)/uniform_random(b) gives a distribution proportional to the denominator, since 1/2=2/4=3/6=... . random() should probably die and randomInt should be rewritten to just forward a random integer from the system random.

DRMacIver:
Interesting. I had exactly the opposite: test/random.clay didn't build on my platform because there were type mismatches! I don't have access to an OSX system, but I'll have a look at the code and see if I can get to the bottom of it.

DRMacIver:
No, my mistake. I was just an idiot and made changes after I tested it. This should work now.

DRMacIver:
And yes, I agree that the implementation looks quite suspicious. I was just picking off the low hanging fruit of it not actually working at all. :-)

jckarter:
Works on my end with your fix too. I pulled the patch.

PR #13, opened 2011-06-05
Platform dependent source file loading

Blei:
This is more or less a copy of how oldclay does it, I don't know if you have other plans to implement this feature (or if you don't want to implement it at all?). So please tell me if you have other suggestions/corrections.

Greetings


jckarter:
I do want to implement a more general conditional loading mechanism, but this will work great as a stopgap. Thanks!

PR #52, opened 2011-12-13
Updated unix bindings for linux64

Blei:
Bindgen now works. Cheers!


jckarter:
Thanks @Blei. Can you also do linux32?

PR #61, opened 2011-12-16
Better runtime error msgs for invalid number parsing

ghost:
Use repr(str) when reporting runtime errors from numbers/parser for
invalid format or overflow.


jckarter:
Merged into master. Thanks!

PR #70, opened 2011-12-19
Collapsed code duplication into 3 generic functions.

ghost:


jckarter:
Looks good. The `_parse*Any` functions should perhaps be private or else given non-underscored names, maybe `parseIntWith`, `parseFloatWith` etc.

Issue #21, opened 2011-12-10
New bindgen needs to generate enum constants with correct type


jckarter:
The new bindgen aliases generated enum types to the correct corresponding C integer type. However, it doesn't emit the corresponding enum constants as the correct type, so they need to be cast when used in conjunction with the enum type.

jckarter:
Since `enum` is often used also to define literal constants in C this will probably be too much of a pain in the ass without #46.

jckarter:
done by @Blei

Issue #71, opened 2011-12-19
inconvenient handling of enums by bindgen


Blei:
The following C header:

``` c
typedef enum {
  FOO = 1,
} my_enum;

void callWithEnum(my_enum e);
```

gets generated as:

```
/* automatically generated by clay-bindgen */

alias Enum_Unnamed1 = UInt;
alias FOO = 1;

alias my_enum = Opaque /* unknown EnumDecl  */;

external (cdecl) callWithEnum(arg0:my_enum) : ;
```

The problem is that `callWithEnum` expects a `my_enum` value as input. This is an alias to `Opaque` which is a `Byte`. The generated enums on the other hand are `Int`s, so type checking fails. It's of course possible to cast all enum values by hand, but it's very inconvenient.

How about something like:

```
alias Enum_Unnamed1 = UInt;
alias FOO = Enum_Unnamed1(1);
alias my_enum = Enum_Unnamed1;
```

Blei:
Oh, I just saw that this is a duplicate of #21. Sorry about that.

jckarter:
It looks like a real bug in there. `my_enum` shouldn't be getting generated as `Opaque`, it should be aliased to the unnamed enum.

PR #72, opened 2011-12-19
bindgen: alias the enum type to the underlying integer type

Blei:
A somewhat hacky fix to #21. Please restructure/rewrite as you see fit :)

### DIFF
--- a/tools/bindgen.clay
+++ b/tools/bindgen.clay
@@ -247,7 +247,8 @@ visitEnum(cursor:CXCursor, parent:CXCursor, data:CXClientData) : Int
     switch (cursor.kind)
     case (CXCursor_EnumConstantDecl) {
         printlnTo(g_bindingsOut, "alias ", clang_getCursorSpelling(cursor), " = ",
-            clang_getEnumConstantDeclValue(cursor), ";");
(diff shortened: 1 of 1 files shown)


jckarter:
@Blei Sorry for taking so long to get back to you on this. I think my fix for #71 fixes this too. Let me know how it goes. If it still sucks feel free to reopen this pull request and I'll take another look.

Blei:
Thank you for your fix. Bindgen behaves somewhat better, but still not quite perfect. The example from #71 now gets generated as:

```
/* automatically generated by clay-bindgen */

alias Enum_Unnamed1 = UInt;
alias FOO = 1;

alias my_enum = Enum_Unnamed1;

external (cdecl) callWithEnum(arg0:my_enum) : ;
```

Which is mostly correct, but the problem of `Int` constants vs. `UInt` enum type still stands:

```
main() {
    callWithEnum(FOO);
}
```

fails with `error: no matching overload found` and `call(CCodePointer[[UInt32], []], Int32)`.

Edit: also, I apparently don't have the necessary rights to reopen this, but I guess you get a notification message anyways :)

jckarter:
@Blei I think setting the correct type for enum constants will break too much stuff, so I want to hold off on that for the meantime.

Issue #82, opened 2011-12-21
Bitwise operators


jckarter:
C programmers love 'em, we should have 'em.
- Change the dereference operator from `a^` to `a@`.
- Add the following desugars:
  - `~a` -> `bitnot(a)`
  - `a & b` -> `bitand(a, b)`
  - `a | b` -> `bitor(a, b)`
  - `a ^ b` -> `bitxor(a, b)`
  - `a << b` -> `bitshl(a, b)`
  - `a >> b` -> `bitshr(a, b)`
  - Rust's precedence rules for these operators is better than C's.

jckarter:
Infix `&`, `|`, `<<`, and `>>` are available, but infix `^` would conflict with the postfix dereference operator. Need to either change the deref operator and use `^` for xor, or pick another random symbol for xor.

ghost:
I'm all for keeping the bitwise operators in-line with C, I think most programmers would be familiar with this syntax.

~ (bitnot)
& (bitand)
| (bitor)
<< (bitshl)

> > (bitshr)
> > ^ (bitxor)

Using the postfix ^ for dereferencing was a big improvement over C pointer syntax, but my preference for dereferencing would just be the array syntax minus the index . . . . ptr[] . . .  would this work?  Failing that some other easily available character, ¬, ?, @, . . .

jckarter:
`[]` is an interesting idea, but it might be too bulky. I was thinking of trying postfix `!` for dereference, at least to start out with.

ghost:
I suppose another issue is <code>&</code> doubling up for address return and bitand. Are you considering changing the address operator? 

Maybe 
<code>ref b = @a</code> 
<code>var c = b!</code>
or
<code>ref b = @a</code> 
<code>var c = b?</code>

As a single character dereferencing operator the <code>?</code> seems fairly intuitive. Although I guess this is being used as a convention for boolean functions. While the <code>!</code> is compact it does seem like a random choice.

jckarter:
Well, the address operator is prefix, so there's no ambiguity between `&a` and `a & b`, like in C.

jckarter:
Actually postfix `!` won't work too because of the `!=` operator—`a! = b;` and `a != b;` would be ambiguous. How about postfix `@`? The pronunciation "at" sounds vaguely pointer-related.

ghost:
Hmm, I can't think of a very good reason not to use that.

[...]

PR #93, opened 2011-12-28
issue #82 - bitwise operators

ghost:


jckarter:
It looks like your original commit is getting cancelled out by your revert commit. You may need to reapply that commit to the branch.

ghost:
I've re-applied the original commit for this.

lib-clay/cocoa/objc/objc.clay
```diff
@@ -59,18 +59,18 @@ overload encodeIdentifier() = #"";
 
 [X] overload encodeIdentifier(static X) = #"?";
 [T] overload encodeIdentifier(static Pointer[T])
-    = IdentifierConcat(#"^", encodeIdentifier(T));
+    = IdentifierConcat(#"@", encodeIdentifier(T));
```
jckarter, line 62: @agemogolk Looks like your update accidentally changed these `^` to `@`, and a few other non-code `^` characters. Maybe try matching a regex like `\b^+` instead of just `^`?

ghost, line 62: Sorry, thought I'd reverted that - probably only did it in my working 
copy though.

ghost:
I've been through the updates for bitwise operators for the current master - and it doesn't work. Something has changed since introducing these changes, haven't figured it out yet so closing this request for now.

jckarter:
Weird. What was the problem exactly?

jckarter:
One thing I recently changed was to make it so that compiler hook functions are implicitly defined in the `__operators__` module now. The problem may be that you need to add lines to loader.cpp to create operator symbols for `bitand`, `bitor`, etc., and on the lib-clay side public import those into `operators` instead of defining them there.

ghost:
Yes, I made those changes and others to bring it up to date, however it 
broke operators in general e.g. building bindgen fell over at the first 
occurrence of "!="

PR #167, opened 2012-01-08
add the <<= operator

aep:
the <<= operator allows async unboxing in the library

"var a <<= f;"  expands to "return yield(f, (a)->
and the rest of the current block is put into the lambda body

inspired by C#'s  await/async
kudos to Joe Groff

### DIFF
--- a/compiler/src/lexer.cpp
+++ b/compiler/src/lexer.cpp
@@ -128,7 +128,7 @@ static void initKeywords() {
          "switch", "case", "break", "continue", "for", "in",
          "true", "false", "try", "catch", "throw",
          "finally", "onerror",
-         "eval",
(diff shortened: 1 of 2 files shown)


jckarter:
Looks cool. I'll throw this in the 0.2 pile, if you don't mind. Thanks!

ghost:
Is this going to conflict with the proposed bitwise operators?

jckarter:
@agemogolk On IRC we talked about changing the syntax into something more in-line with the other variable binding forms, for instance `bind x = async();` instead of `var x <<= async();`.

aep:
closing to make a new one (github is odd)

aep:
https://github.com/jckarter/clay/pull/184

PR #177, opened 2012-01-10
operators: fixed borked comparisons

Blei:


ghost:
Ooops on my part, forgot to change the operator module back part way through testing. The changes should be reverted back to the originals rather the fix given here (fixed in my latest master commit).

Blei:
ah, thanks. closing this.

PR #180, opened 2012-01-12
add dependency file generation akin to -MMD on gcc

Blei:
I'm not sure if I violated some code guidelines/best practices/clay coding conventions (I'm not much of a C++ programmer usually). Also, this probably still contains bugs. Please don't hesitate to correct stuff or send complaints :)

### DIFF
--- a/compiler/src/clay.hpp
+++ b/compiler/src/clay.hpp
@@ -2341,7 +2341,7 @@ extern map<string, string> globalFlags;
 extern ModulePtr globalMainModule;
 
 void addSearchPath(const string &path);
-ModulePtr loadProgram(const string &fileName);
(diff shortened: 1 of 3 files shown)


jckarter:
Looks good at a glance. Have you tried it with circular imports or any other corner cases just to make sure it doesn't blow up?

Blei:
It only includes files when these same files are loaded by the existing code, so everything should work out all right.

At the moment, the list of dependencies is conservative, in the sense that all loaded files are added to it. It might be possible to prune this list to only include the files of which code/variables are actually used.

jckarter:
Well runtests with --buildflags=-deps didn't crash, so that means it works, right? Looks good.

PR #196, opened 2012-01-21
WindowsError is subclass of OSError

ghost:
Catching just OSError should work.


ghost:
Doh - this is already in jckarter/clay@959434a

PR #204, opened 2012-01-27
with/monad

aep:


galchinsky:
How should I write main() with `with` block?

http://pastebin.com/hVMjJaH1

Haskell's do-notation seems more clear to me.

PR #206, opened 2012-01-29
Updates to math module - libm default for non-complex functions - issue#153

ghost:
Have set all non-complex math functions to default to libm. The utility functions (nan?, copysign, etc) are using native versions implemented in numbers/floats module where possible.


jckarter:
Is this branch based off of master? There's a bunch of stuff in the diff that's already in master.

Issue #208, opened 2012-02-03
[patch] hashmaps should grow occasionally


mrjbq7:
I'm not much of a Clay programmer, but this patch should make hashmaps like a million times better... what do you think?

``` diff
diff --git a/lib-clay/hashmaps/hashmaps.clay b/lib-clay/hashmaps/hashmaps.clay
index 5c0bdae..038413e 100644
--- a/lib-clay/hashmaps/hashmaps.clay
+++ b/lib-clay/hashmaps/hashmaps.clay
@@ -84,6 +84,9 @@ overload put(a:HashMap[K,V], key:K, forward value:V) {
     ref bucket = a.buckets[i];
     bucket = UniqueValue(Node(key, value, move(bucket)));
     a.size += 1;
+    if (a.size > (size(a.buckets) * 0.75)) {
+        grow(a);
+    }
 }


@@ -99,9 +102,13 @@ overload index(a:HashMap[K,V], key:K) {
     if (not null?(ptr))
         return ref ptr^;
     ref bucket = a.buckets[i];
-    bucket = UniqueValue(Node(key, V(), move(bucket)));
+    var value = V();
+    bucket = UniqueValue(Node(key, value, move(bucket)));
     a.size += 1;
-    return ref bucket^.value;
+    if (a.size > (size(a.buckets) * 0.75)) {
+        grow(a);
+    }
+    return ref value;
 }


@@ -213,3 +220,29 @@ private findInBucket(_node:NodePtr[K,V], key:K) {
     }
     return null(V);
 }
+
+[K,V]
+private grow(a:HashMap[K,V]) {
+    var oldbuckets = a.buckets;
+
+    a.buckets <-- Vector[NodePtr[K,V]]();
+    a.size = 0;
+
+    try {
+        resize(a.buckets, size(oldbuckets) * 2);
+    } catch (ex) {
+        destroy(a.buckets);
+        throw ex;
+    }
+
+    for(i in range(0, size(oldbuckets))) {
+        var node = &oldbuckets[i]^;
+        while (not null?(node)) {
+            put(a, node^.key, node^.value);
+            node = &node^.next^;
+        }
+    }
+
+    destroy(&oldbuckets);
+}
```

mrjbq7:
Closing this since there's a pull request now...

PR #209, opened 2012-02-03
hashmaps: grow buckets when exceeding 75% load factor.

mrjbq7:
See issue #208.

### DIFF
--- a/lib-clay/hashmaps/hashmaps.clay
+++ b/lib-clay/hashmaps/hashmaps.clay
@@ -101,14 +101,14 @@ overload index(a:HashMap[K,V], key:K) {
     var ptr = findInBucket(a.buckets[i], key);
     if (not null?(ptr))
         return ref ptr^;
-    ref bucket = a.buckets[i];
-    var value = V();
-    bucket = UniqueValue(Node(key, value, move(bucket)));
-    a.size += 1;
-    if (a.size > (size(a.buckets) * 3 / 4)) {
+    if (a.size >= (size(a.buckets) * 3 / 4)) {
         grow(a);
+        i = bucketId(a, key); // recalculate index
     }
-    return ref value;
+    ref bucket = a.buckets[i];
+    bucket = UniqueValue(Node(key, V(), move(bucket)));
+    a.size += 1;
+    return ref bucket^.value;
 }
 
 
--- a/lib-clay/hashmaps/hashmaps.clay
+++ b/lib-clay/hashmaps/hashmaps.clay
@@ -233,7 +233,7 @@ private grow(a:HashMap[K,V]) {
     for(i in range(0, size(oldbuckets))) {
         var node = &oldbuckets[i]^;
         while (not null?(node)) {
-            put(a, node^.key, node^.value);
+            put(a, move(node^.key), move(node^.value));
             node = &node^.next^;
         }
     }
--- a/lib-clay/hashmaps/hashmaps.clay
+++ b/lib-clay/hashmaps/hashmaps.clay
@@ -223,17 +223,12 @@ private findInBucket(_node:NodePtr[K,V], key:K) {
 
 [K,V]
 private grow(a:HashMap[K,V]) {
-    var oldbuckets = a.buckets;
+    var oldbuckets = move(a.buckets);
 
     a.buckets <-- Vector[NodePtr[K,V]]();
     a.size = 0;
 
-    try {
-        resize(a.buckets, size(oldbuckets) * 2);
-    } catch (ex) {
-        destroy(a.buckets);
-        throw ex;
-    }
+    resize(a.buckets, size(oldbuckets) * 2);
 
     for(i in range(0, size(oldbuckets))) {
         var node = &oldbuckets[i]^;
@@ -242,6 +237,4 @@ private grow(a:HashMap[K,V]) {
             node = &node^.next^;
         }
     }
-
-    destroy(&oldbuckets);
 }
--- a/lib-clay/hashmaps/hashmaps.clay
+++ b/lib-clay/hashmaps/hashmaps.clay
@@ -84,6 +84,9 @@ overload put(a:HashMap[K,V], key:K, forward value:V) {
     ref bucket = a.buckets[i];
     bucket = UniqueValue(Node(key, value, move(bucket)));
     a.size += 1;
+    if (a.size > (size(a.buckets) * 0.75)) {
+        grow(a);
+    }
 }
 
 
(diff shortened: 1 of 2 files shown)


lib-clay/hashmaps/hashmaps.clay
```diff
@@ -84,6 +84,9 @@ overload put(a:HashMap[K,V], key:K, forward value:V) {
     ref bucket = a.buckets[i];
     bucket = UniqueValue(Node(key, value, move(bucket)));
     a.size += 1;
+    if (a.size > (size(a.buckets) * 0.75)) {
```
jckarter: `* 3 / 4` would be better here; no point converting to float and back

lib-clay/hashmaps/hashmaps.clay
```diff
@@ -99,9 +102,13 @@ overload index(a:HashMap[K,V], key:K) {
     if (not null?(ptr))
         return ref ptr^;
     ref bucket = a.buckets[i];
-    bucket = UniqueValue(Node(key, V(), move(bucket)));
+    var value = V();
+    bucket = UniqueValue(Node(key, value, move(bucket)));
     a.size += 1;
-    return ref bucket^.value;
+    if (a.size > (size(a.buckets) * 0.75)) {
```
jckarter: same here

mrjbq7:
Okay, how's that?

lib-clay/hashmaps/hashmaps.clay
```diff
@@ -213,3 +220,28 @@ private findInBucket(_node:NodePtr[K,V], key:K) {
     }
     return null(V);
 }
+
+[K,V]
+private grow(a:HashMap[K,V]) {
+    var oldbuckets = a.buckets;
+
+    a.buckets <-- Vector[NodePtr[K,V]]();
```
jckarter, line 228: This will leak memory. `var oldbuckets = a.buckets;` copies the old vector, then `a.buckets <-- Vector[...]();` overwrites `a.buckets` without releasing the memory. You can use `move` to safely move `a.buckets` into `oldbuckets`; after moving, `a.buckets` will be empty. Try this:

```
[K,V]
private grow(a:HashMap[K,V]) {
    var oldbuckets = move(a.buckets);

    a.size = 0;

    ...
}
```

lib-clay/hashmaps/hashmaps.clay
```diff
@@ -213,3 +220,28 @@ private findInBucket(_node:NodePtr[K,V], key:K) {
...
+    a.size = 0;
+
+    try {
+        resize(a.buckets, size(oldbuckets) * 2);
+    } catch (ex) {
+        destroy(a.buckets);
+        throw ex;
+    }
```
jckarter: Remove this `try` block; there's no need to manually destroy `a.buckets` if you use `move` like i described on line 228.

lib-clay/hashmaps/hashmaps.clay
```diff
@@ -213,3 +220,28 @@ private findInBucket(_node:NodePtr[K,V], key:K) {
...
+
+    for(i in range(0, size(oldbuckets))) {
+        var node = &oldbuckets[i]^;
+        while (not null?(node)) {
+            put(a, node^.key, node^.value);
+            node = &node^.next^;
+        }
+    }
+
+    destroy(&oldbuckets);
```
jckarter: You don't need to manually destroy local variables.

lib-clay/hashmaps/hashmaps.clay
mrjbq7: I'm trying to grow the a.buckets here to be twice as large -- using the convention from line 28 in `hashmaps.clay` where it tries to resize and destroys if it fails?  Is that not correct?

lib-clay/hashmaps/hashmaps.clay
mrjbq7: How do I know if my program leaks?  Is there valgrind support?

jckarter: Yes. Run a clay program under valgrind and it should just work.

lib-clay/hashmaps/hashmaps.clay
jckarter: Line 28 is in the constructor and uses named return values. Because `returned` isn't a fully initialized value until the function returns, any partially initialized components must be manually destroyed if the constructor fails. If you don't use named return values or raw allocation you never need to use `<--` or manually call `destroy`.

mrjbq7: So I can just `resize(a.buckets, size(oldbuckets) * 2)` and ignore any exception?

jckarter: Yeah. If resize throws then the hashtable will be hosed, but all memory will at least be cleaned up.

`grow` could maybe be made more exception-robust if you initialized `newbuckets` outside of the object then did `swap(a.buckets, newbuckets);` at the end.

mrjbq7:
Okay, updated the pull request with your suggestions...

lib-clay/hashmaps/hashmaps.clay
```diff
@@ -99,9 +102,13 @@ overload index(a:HashMap[K,V], key:K) {
...
     ref bucket = a.buckets[i];
-    bucket = UniqueValue(Node(key, V(), move(bucket)));
+    var value = V();
+    bucket = UniqueValue(Node(key, value, move(bucket)));
     a.size += 1;
-    return ref bucket^.value;
+    if (a.size > (size(a.buckets) * 3 / 4)) {
+        grow(a);
+    }
+    return ref value;
```
jckarter: This doesn't look right. You want to return a reference to the value inside the bucket, not the local variable. You should revert these changes other than the growth check.

mrjbq7: Does bucket still point to the old bucket in the old list after grow() ? I wasn't sure.

jckarter: No, since you allocate a new buckets vector, the `bucket` ref will be invalid. You'll need to `return ref a.buckets[i]^.value;` instead.

mrjbq7: but `a.buckets[i]` is wrong since the index might have changed after rehash?  I was trying to hold onto the old location in memory or cache the old value somehow...

jckarter: If the index changes, then you'll need to recalculate it. You can use the saved `bucket` reference in the case when it doesn't grow.

lib-clay/hashmaps/hashmaps.clay
```diff
@@ -213,3 +220,21 @@ private findInBucket(_node:NodePtr[K,V], key:K) {
...
+
+    a.buckets <-- Vector[NodePtr[K,V]]();
+    a.size = 0;
+
+    resize(a.buckets, size(oldbuckets) * 2);
+
+    for(i in range(0, size(oldbuckets))) {
+        var node = &oldbuckets[i]^;
+        while (not null?(node)) {
+            put(a, node^.key, node^.value);
```
jckarter: Since you're transferring values from the old bucket, you could move the key and value with `put(a, move(node^.key), move(node^.value));` here.

lib-clay/hashmaps/hashmaps.clay
jckarter: It looks like you could refactor `index` so that it `grow`s first if necessary then proceeds as it does currently. Does that seem reasonable? Then you wouldn't have any redundant calculation of the new key's bucket.

mrjbq7:
Okay, incorporated that feedback.

PR #211, opened 2012-02-10
Factor out default constructors

### LINKED ISSUE #187: Factor out default constructor implementations
The default constructor implementations should be available from the library under other names, such as `initializeRecord` `initializeVariant` etc., so that `T()` can be overloaded without losing access to the base implementation.

Blei:
Fixes #187

### DIFF
--- a/lib-clay/records/records.clay
+++ b/lib-clay/records/records.clay
@@ -66,9 +66,7 @@ alias initializeRecord(static T, ..args:A) --> returned:T {
 }
 
 [T, ..A | Record?(T) and (Tuple[..RecordFieldTypes(T)] == Tuple[..A])]
-alias overload T(..args:A) --> returned:T {
-    returned <-- initializeRecord(T, ..args);
-}
+alias overload T(..args:A) = initializeRecord(T, ..args);
 
 
 
@@ -76,15 +74,13 @@ alias overload T(..args:A) --> returned:T {
 // default constructor
 //
 
-[T | RegularRecord?(T)]
+[T | Record?(T)]
 overload initializeRecord(static T) --> returned:T {
     initializeMulti(..recordFields(returned));
 }
 
 [T | RegularRecord?(T)]
-overload T() --> returned:T {
-    returned <-- initializeRecord(T);
-}
+overload T() = initializeRecord(T);
 
 [..A]
 private initializeMulti(..args:A) {
@@ -102,9 +98,12 @@ inline overload initializeTypes() = ;
 //
 
 [T | RegularRecord?(T) and not BitwiseCopiedType?(T)]
-overload T(src:T) --> returned:T {
+copyRecord(src:T) --> returned:T {
     ..recordFields(returned) <-- ..recordFields(src);
 }
+
+[T | RegularRecord?(T) and not BitwiseCopiedType?(T)]
+overload T(src:T) = copyRecord(src);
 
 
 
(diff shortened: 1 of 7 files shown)


lib-clay/records/records.clay
```diff
@@ -72,10 +77,15 @@ alias overload T(..args:A) --> returned:T {
 //
 
 [T | RegularRecord?(T)]
```
jckarter: The predicate on `initializeRecord` should be `Record?`, not `RegularRecord?`.

lib-clay/records/records.clay
```diff
@@ -72,10 +77,15 @@ alias overload T(..args:A) --> returned:T {
...
-overload T() --> returned:T {
+overload initializeRecord(static T) --> returned:T {
     initializeMulti(..recordFields(returned));
 }
 
+[T | RegularRecord?(T)]
+overload T() --> returned:T {
+    returned <-- initializeRecord(T);
+}
```
jckarter: This could just be `overload T() = initializeRecord(T);`

lib-clay/records/records.clay
```diff
@@ -61,10 +61,15 @@ overload RegularRecord?(static T) = true;
...
 [T, ..A | Record?(T) and (Tuple[..RecordFieldTypes(T)] == Tuple[..A])]
-alias overload T(..args:A) --> returned:T {
+alias initializeRecord(static T, ..args:A) --> returned:T {
     ..recordFields(returned) <-- ..args;
 }
 
+[T, ..A | Record?(T) and (Tuple[..RecordFieldTypes(T)] == Tuple[..A])]
+alias overload T(..args:A) --> returned:T {
+    returned <-- initializeRecord(T, ..args);
+}
```
jckarter: `alias overload T(..args:A) = initializeRecord(T, ..args);`

lib-clay/variants/variants.clay
```diff
@@ -28,13 +28,18 @@ VariantMember?(static V, static M) =
...
     ref repr = variantRepr(returned);
     alias index = Int(VariantMemberIndex(V, T));
     repr.tag <-- index;
     staticIndex(repr.data, static index) <-- x;
 }
 
+[V,T | Variant?(V) and VariantMember?(V, T)]
+alias overload V(x:T) --> returned:V {
+    returned <-- initializeVariant(V, x);
+}
```
jckarter: `alias overload V(x:T) = initializeVariant(V, x);`

jckarter:
Thanks! It would also be useful to have `copyRecord`, `moveUnsafeRecord`, `resetUnsafeRecord`, and `destroyRecord` factored out as well.

Blei:
Implemented and rebased.

lib-clay/records/records.clay
```diff
@@ -92,10 +98,13 @@ inline overload initializeTypes() = ;
 //
 
 [T | RegularRecord?(T) and not BitwiseCopiedType?(T)]
-overload T(src:T) --> returned:T {
+copyRecord(src:T) --> returned:T {
```
jckarter, line 101: The predicate on `copyRecord` and the other factors should use `Record?` instead of `RegularRecord?`. Looks good otherwise.

lib-clay/variants/variants.clay
jckarter: You should still simplify this to `alias overload V(x:T) = initializeVariant`

lib-clay/records/records.clay
jckarter, line 101: Oh, also, `copyRecord` etc. shouldn't be predicated on `BitwiseCopiedType?` etc. either. They should be usable with any record type.

Blei:
Done.

Issue #46, opened 2011-12-12
Use C conversion rules in call(CCodePointer, ..)


jckarter:
C functions are monomorphic and C APIs often have weak integer type discipline, so it would be convenient for `call(CCodePointer, ..)` to perform the same implicit conversions as C automatically.

jckarter:
Fixed by @Blei In #212

PR #212, opened 2012-02-10
Automatic integer casts when calling external functions for some cases

Blei:
(Partial) fix for #46

See comments in the code.


Blei:
I implemented your suggestions and rebased to the current master.

jckarter:
Thanks. Now that this is done, do you still have your bindgen patch for converting enum values as the correct type?

Blei:
Yes, just rebased it: https://github.com/Blei/clay/tree/bindgen-enums

PR #213, opened 2012-02-10
Bitwise ops issue#82

ghost:
Alterations as discussed in issue #82
All libs updated.
All tests updated.
Bitwise ops tests added.


lib-clay/math/native/pow_std.clay
```diff
@@ -3110,7 +3110,7 @@ var  twoPowerIndex2 = Array[tPi2_t,256](
    Decomposes a positive double precision
    variable x (normal or subnormal) such that
 
-   2^(resE) * resm = x
+   2@(resE) * resm = x
```
jckarter, line 3113: It looks like you still accidentally transformed all these `^` into `@` in math comments.

lib-clay/win32/errno/errno.clay
```diff
@@ -1,6 +1,6 @@
 import libc;
 
-errno() = libc._errno()^;
+errno() = win32._errno()@;
```
jckarter, line 3: Looks like you accidentally reverted the win32 -> libc change here

test/cocoa/objc/internals/test.macosx.clay
```diff
@@ -251,19 +251,19 @@ main() = testMain(TestSuite("Objective-C binding internals", array(
...
+        expectEncoding(test, "@@{Foo}", Pointer[Pointer[Foo]]);
+        expectEncoding(test, "@@@{Foo}", Pointer[Pointer[Pointer[Foo]]]);
         expectEncoding(test, "{Bar=q{Foo=@*i}}", Bar);
         expectEncoding(test, "(Union=i{Foo=@*i}{Bar=q{Foo=@*i}})", Foonion);
-        expectEncoding(test, "^(Union=i{Foo=@*i}{Bar=q{Foo=@*i}})", Pointer[Foonion]);
-        expectEncoding(test, "^^(Union)", Pointer[Pointer[Foonion]]);
-        expectEncoding(test, "^^^(Union)", Pointer[Pointer[Pointer[Foonion]]]);
+        expectEncoding(test, "@(Union=i{Foo=@*i}{Bar=q{Foo=@*i}})", Pointer[Foonion]);
+        expectEncoding(test, "@@(Union)", Pointer[Pointer[Foonion]]);
+        expectEncoding(test, "@@@(Union)", Pointer[Pointer[Pointer[Foonion]]]);
```
jckarter, line 266: these objc encoding strings shouldn't be changed

test/random/main.clay
```diff
@@ -2,7 +2,7 @@ import test.*;
 import random.*;
 
 // WARNING: This test suite may fail generately. Care is taken to assure that the probability of it doing so 
-// is relatively low in the case that there is a genuine failure (<= 2^-64 is the general goal).
+// is relatively low in the case that there is a genuine failure (<= 2@-64 is the general goal).
```
jckarter, line 5: comment shouldn't be changed

test/unmaintained/sieve.clay
```diff
@@ -0,0 +1,95 @@
```
jckarter, line 0: You added back this file which should be removed. Perhaps you should rebase against master?

jckarter:
You said you added tests for the bitwise ops, but I don't see any new tests in the diff. Did you forget to add them?

ghost:
The new tests are in the operators/bitwise test.

jckarter:
Another syntactic issue is that `|` is also used in pattern guard syntax. While it's not ambiguous, it's kind of icky. Any ideas for a better token to put there?

ghost:
OK, i've re-based the bitwise-ops branch and think I have all issues fixed.

Not sure about a different symbol for bitor, seems like all the other single symbols are used.

jckarter:
I was actually thinking of a different symbol for the pattern guard. `[T if Foo?(T)]` is an obvious contender, though `if` is also already valid expression syntax too.

ghost:
How about 'where' as a pattern guard?  Too verbose?

ghost:
Is '::' already used somewhere ? Actually I think 'when' would be better 
than 'where' . . .

jckarter:
`where` is a bit large, but it's probably better than punning an existing keyword.

jckarter:
`::` is used in call-with-block syntax, for instance, `foo() : x -> { } :: y -> { }`, calls foo with the lambdas `x -> { }` and `y -> { }`. `when` seems like a good choice.

ghost:
I've always thought pattern guards looked like a lambda function 
specialised for predicates. How about using a similar (same?) token 
rather than introduce another keyword/operator? Maybe something like 
'~>' , '->?' or '><'.

While we're on operators how about a concatenation operator?  I realise 
'+' is the common string concat, but it doesn't seem right when applied 
to a sequence of any type (I guess I prefer my 'add' functions to do 
arithmetic only). Maybe '++' or '&&' or whatever.

jckarter:
On Sat, Feb 11, 2012 at 4:35 PM, agemogolk
reply@reply.github.com
wrote:

> I've always thought pattern guards looked like a lambda function
> specialised for predicates. How about using a similar (same?) token
> rather than introduce another keyword/operator? Maybe something like
> '~>' , '->?' or '><'.

I'm not a fan of weird symbols. I'd prefer a keyword.

> While we're on operators how about a concatenation operator?  I realise
> '+' is the common string concat, but it doesn't seem right when applied
> to a sequence of any type (I guess I prefer my 'add' functions to do
> arithmetic only). Maybe '++' or '&&' or whatever.

Getting rid of `+` as concatenation is a good idea. I like `++`. It
would be nice if concatenation sequences desugared as a single
invocation, so that `a ++ b ++ c` became `cat(a, b, c)` instead of
`cat(cat(a, b), c)`, to avoid the overhead typically incurred by C++
implementations of concatenation operators.

-Joe

ghost:
For a short keyword how about 'si', I think that translates to 'if' in a 
few languages . . .

Maybe raise the 'cat' operator as an issue and i'll take a look at it.

jckarter:
On Sat, Feb 11, 2012 at 4:48 PM, agemogolk
reply@reply.github.com
wrote:

> For a short keyword how about 'si', I think that translates to 'if' in a
> few languages . . .

`when` or `where` seemed like fine candidates.

ghost:
Wasn't serious about the 'si'!

Always better to stick with English (or in some cases i suppose American 
will do too . . ).

PR #214, opened 2012-02-10
Bitwise ops issue#82

ghost:
Think this is actually ready to go now.


ghost:
Bitwise operators definable outside of compiler so this branch is no longer valid - issue#240

Issue #215, opened 2012-02-12
Don't use `+` for concatenation


jckarter:
Using `+` and `+=` for sequence concatenation is dumb. `push` can be used in place of `+=`, and a variadic `cat()` function would be better than `+`.

jckarter:
Closed by @agemogolk in #216

PR #216, opened 2012-02-18
Concatenation operator "++" -  issue #215

ghost:
Introduced a "variadic operator" which desugars into variadic function call.
Default sequences concatenation updated.
All tests updated.


jckarter:
Looks great! I think we could promote the arithmetic binary ops to variadic ops as well; `add` and `multiply` are already overloaded for variadic argument lists, and it would be easy to do the same for `subtract`, `divide`, and `remainder`.

I was thinking it'd also be cool to support Python-style chained comparisons (such as `3. <= a < 5.` for `3. <= a and a < 5.`).  What do you think of that?

jckarter:
Another thing: for consistency, there should probably also be a `++=` update operator.

ghost:
> Where are  `add` and `multiply`  overloaded for variadic args?

Looks like I'm mistaken about that. I thought they were. At any rate, they could be extended in the same way `bitor` and `bitand` are for variadic arguments.

PR #217, opened 2012-02-18
Concatenation update-assign operator "++="

ghost:


jckarter:
Looks good. Do you also want to implement the library end by overloading `updateAssign`? It could likely just forward to `push`.

ghost:
Oh right, forgot about that. I'll update it.

PR #218, opened 2012-02-18
Update assign for cat

ghost:
var a = String("a");
a ++= "b";
updateAssign(cat,a,"c","d","e");
println(a);

out:
abcde


jckarter:
Could you add your example as a test case?

PR #221, opened 2012-02-19
Promotion of binary-ops to variadic-ops.

ghost:
Outstanding issues:
    - Pointer arithmetic requires parens
    - all ops have precedence *,/,%,-,+
    - io/paths test fails with subtractOverflow - dunno why.

Arith operators overloaded for variadic args.

### DIFF
--- a/lib-clay/numbers/numbers.clay
+++ b/lib-clay/numbers/numbers.clay
@@ -290,20 +290,20 @@ inline overload remainder(a:A, b:A) = integerRemainder(a, b);
 // Variadic ops
 //
 
-[A, ..AA | allValues?(Numeric?,A, ..AA) and countValues(..AA) != 1]
-inline overload add(a:A, ..aa:AA) : A = foldValues(add, a, ..aa);
+[A, ..AA | countValues(..AA) != 1]
+inline overload add(a:A, ..aa:AA) : A = foldlValues(add, a, ..aa);
 
-[A, ..AA | allValues?(Numeric?,A, ..AA) and countValues(..AA) != 1]
-inline overload subtract(a:A, ..aa:AA) : A = foldValues(subtract, a, ..aa);
+[A, ..AA | countValues(..AA) != 1]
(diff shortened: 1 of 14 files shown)


lib-clay/numbers/numbers.clay
```diff
@@ -287,6 +287,25 @@ inline overload remainder(a:A, b:B) =
...
+inline overload subtract(a:A, ..aa:AA) : A = foldValues(subtract, a, ..aa);
+
+[A, ..AA | allValues?(Numeric?,A,..AA) and countValues(..AA) != 1]
+inline overload multiply(a:A, ..aa:AA) : A = foldValues(multiply, a, ..aa);
+
+[A, ..AA | allValues?(Numeric?,A,..AA) and countValues(..AA) != 1]
+inline overload divide(a:A, ..aa:AA) : A = foldValues(divide, a, ..aa);
+
+[A, ..AA | allValues?(Integer?,A,..AA) and countValues(..AA) != 1]
+inline overload remainder(a:A, ..aa:AA) : A = foldValues(remainder, a, ..aa);
```
jckarter: Instead of predicating the variadic forms on `Numeric?` or `Integer?`, I think the predicate here should merely be `countValues(..AA) != 1`. I think you should also use `foldlValues` instead of `foldValues`, since you want the operators to be left-associative by default.  Those changes should make the variadic operators drop-in compatible and let you back out your other changes.

lib-clay/numbers/numbers.clay
```diff
@@ -498,12 +517,21 @@ inline wrapDivide(a:A, b:A) : A = numericDivide(a, b);
 [A | Integer?(A)]
 inline wrapRemainder(a:A, b:A) : A = integerRemainder(a, b);
 
-[A, ..AA | equalValues?(A, ..AA) and countValues(..AA) != 1]
```
jckarter, line 501: The `equalValues?` predicate is correct for the `wrap*` functions, since they don't do any type conversion.

ghost, line 501: OK, ta.

PR #222, opened 2012-02-19
Promotion of binary-ops to variadic-ops.

ghost:


lib-clay/deques/deques.clay
```diff
@@ -79,7 +79,7 @@ record DequeCoordinate[T] (
 }
 
 [T] overload subtract(i: DequeCoordinate[T], j: DequeCoordinate[T]) : PtrInt {
-    var mapDistance = PtrInt(i.node - j.node - 1);
+    var mapDistance = PtrInt((i.node - j.node) - 1);
```
jckarter: Can this change be backed out now?

ghost: Hmm, no. I think the problem here is all variadic args are expected to 
be of the same type. Not sure of the best way to fix this . .

jckarter: The type of the args shouldn't matter, since the variadic overload just forwards to the binary overload.

ghost: Ha, daft copy/paste error - left the return type in the arithmetic 
overloads. Think this will solve it.

PR #225, opened 2012-02-21
Use variadicOp desugar to remove superfluous ifExpr code.

ghost:
Move IF_EXPR to variadic op.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -56,6 +56,22 @@ Language changes:
 
   Thanks to Jeremy Clifford.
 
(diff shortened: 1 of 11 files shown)


jckarter:
I pushed a fix for the shootout/mean test too; could you rebase your other patch against master?

ghost:
Done

jckarter:
Thanks. Maybe we could trim even more boilerplate by using the same `VariadicOp` node for all desugared operator nodes, whether unary, binary, or variadic.

PR #227, opened 2012-02-21
Python style chained comparison operators.

ghost:


lib-clay/operators/operators.clay
```diff
@@ -247,6 +248,10 @@ inline overload greater?(a, b) = lesser?(b, a);
 
 inline overload greaterEquals?(a, b) = not lesser?(a, b);
 
+inline overload compare?(expr1, fn, expr2, ..args) : Bool
+    = if(not compare?(expr1, fn, expr2)) false else compare?(expr2, ..args);
```
jckarter: This could be simplified a bit to `fn(expr1, expr2) and compare?(expr2, ..args)`.

ghost: That's how I originally had it,  but then all comparisons in the chain 
are evaluated.

jckarter: Are you sure about that? `and` short circuits.

ghost: Hmmm, forgot about that. I'll change it back.

PR #229, opened 2012-02-29
math lib update + extras  -  fixes issue#153 & issue#178

ghost:
Math libm completely revamped. Based on openlibm. (crlibm ports problematic).

Double precision native functions implemented.
Float/LongDouble mostly still handled by libm.
Compiler flag -Dmath.Mode=libm uses all available libm functions.
Tests added for libm wrappers, native/complex double functions.
- tests take long time due to size of math lib.

Native math lib/tests are comprehensive for Double. Float and LongDouble will be added gradually.

Extras:
Hacked up a the hires-timer for unix in the time module.
Integer div fix in mandelbrot shootout


lib-clay/math/native/sqrt.x86.clay
```diff
@@ -0,0 +1,10 @@
+
+public import math.native.*;
+import simd.*;
+
+private external ("llvm.x86.sse2.sqrt.sd")
```
jckarter: LLVM intrinsic externals need the `llvm` attribute so they will be codegenned as LLVM intrinsics rather than as C functions:

```
private external ("llvm.x86.sse2.sqrt.sd", llvm)
    x86_sqrtsd(a:Vec[Double,2]) : Vec[Double,2];
```

The externals in math/simd/simd.x86.clay should use the `llvm` attribute too.

ghost: Ah, would that explain the 'module broken' error when trying to compile 
the sqrt.x86.clay module with debugging symbols ?

jckarter:
Looks good aside from the intrinsic attributes. If you test on x86-32 rather than -64 it will likely crash if the intrinsics aren't marked properly as `llvm`.

lib-clay/math/native/sqrt.x86.clay
jckarter: Could be, though I would expect LLVM to complain under any circumstance an intrinsic is improperly codegenned not contingent on `-g`.

jckarter:
The math tests are killing my laptop. It looks like the compiler's guzzling 1.6GB of VM a piece to build them. I think you might need to break them down into smaller groups; the `math/complex` test ran fine with its relatively small set of tested functions, so maybe breaking the scalar tests into groups would make compiling them less pathological. The common definitions in the tests could be factored out into a module all the tests import, and then each test could instantiate a different subset of all of the functions.

I'm also seeing some formatting-related failures on Mac OS X; it looks like OS X's printf doesn't print the sign for nan or inf, even with the `+` format flag. An example:

``` diff
-sin(0+nanj) = 0+nanj
-sin(-0+nanj) = 0+nanj
+sin(0nanj) = 0nanj
+sin(-0nanj) = 0nanj
```

jckarter:
A possibly simpler refactoring you could try to improve the test performance: Change the multiple-value lists to arrays and the `..for` loops to normal for loops. That should reduce the raw amount of code the compiler has to generate then chew through. You could pack a set of functions into an array of code pointers using something like this:

```
unaryOps(static T) = array(..mapValues(CodePointer[[T],[T]], sin, cos, /*etc*/));
```

jckarter:
I have a patch to make `nan` print consistently across platforms that should fix the output formatting issue. The tests look correct on OS X with that patch applied (once they finally finish).

ghost:
Fixed up the imports in the math lib and changed the test inner loop  - 
big performance improvement. I'll have a look at your other suggestions 
when I get a chance.

jckarter:
Performance is much better, thanks! The only thing left is that the out.txt for the math tests need to be updated for my `nan` printing patch. I made it so that nan values or complex values with nan components just print as `nan` regardless of sign(s), to avoid dealing with platform-dependent printf behavior.

ghost:
Updated test output to use unsigned nans.

jckarter:
I changed the math/libm test not to check output, since MacOS X's libm gives slightly different results for a bunch of functions, and the important thing with libm is that the function references all compile, not what result they give.

PR #236, opened 2012-03-08
user defined operators

ghost:
Parse all arithmetic as a single variadic expression and handle precedence in operator module.

User defined operators from limited set of symbols (!,%,/,\,+,-,*,<,>,=)


lib-clay/deques/deques.clay
```diff
@@ -44,11 +44,11 @@ record DequeCoordinate[T] (
     dec(i.current);
 }
 
-[T, I | Integer?(I)] overload updateAssign(static add, ref i: DequeCoordinate[T], n: I) {
+[T, I | Integer?(I)] overload updateAssign(static arith, ref i: DequeCoordinate[T],static add, n: I) {
```
jckarter: There should be a default overload for `updateAssign(arith)` so that overloading `updateAssign(add)` etc. still work as before. I think `arith` should be an implementation detail unless you actually need to do something unusual with precedence or fixity.

lib-clay/operators/operators.clay
```diff
@@ -329,20 +330,41 @@ inline overload indexUpdateAssign(static F, dest, forward ..indicesAndSrc:A) {
...
 
+//
+//  default arithmetic - use left associativity with precedence *,/,\,%,-,+ 
+//
+
+overload arith(forward ..args) 
+    = arithp(add,..arithp(subtract,..arithp(remainder,..arithp(quotient,
+        ..arithp(divide,..arithp(multiply,..args))))));
+
+private define arithp;
```
jckarter: This would more idiomatically be named `arith?`

lib-clay/sequences/sequences.clay
```diff
@@ -118,3 +118,55 @@ interleave(seq:S, betweenFn, fn) {
...
+overload seqType(T) = T;
+
+private define seqSize;
+
+overload seqSize(s,a,..args) = ..seqSize(s,..args);
+[S | Sequence?(S)] 
+overload seqSize(s,a:S,..args) = ..seqSize(max(s,size(a)),..args);
+overload seqSize(s) = s;
+
+private define reduceToScaler;
```
jckarter: Did you mean `reduceToScalar`?

jckarter:
This is cool. It would be nice to have some tests that show that the default `arith` overload provides the correct traditional precedence for the math operators.

lib-clay/sequences/sequences.clay
ghost: lol yeah, find/replace a typo!

lib-clay/operators/operators.clay
ghost: Not really happy with 'arith' just couldn't come up with anything better at the time.

ghost:
OK, i'll put some tests together, wanted to see if it was worth doing first.

lib-clay/operators/operators.clay
jckarter: Do `arith` and `compare` need to be different functions? Maybe they could be combined into a single `operator` function.

jckarter:
If the add, subtract, etc. functions aren't needed for syntax expansion anymore, then they could be removed from the compiler's magic `__operators__` module and defined as normal symbols in lib-clay's operators module.

ghost:
Not entirely sure about this one, can you just set the de-sugered 
callable to a clay expression rather than operator_expr_\*  . . . ?

I've made all other adjustments and merged arith/cat/compare. Some of it 
looks a little clunky (expanded operator cases and sequences operator 
overload pattern), but it seems to work ok.

I may get around to doing some tests tomorrow.

jckarter:
You're right, you still need the `operator_expr_*` symbols for the expansion. Never mind.

jckarter:
If you want to take it another step, a cool idea would be to make the operators user-defined like in ML or Haskell: allow any string of designated operator characters to be parsed as an operator, and pass the operator name as a static string to the `operator` function. For example, `1 + 2 / 3` could call `operator(1, #"+", 2, #"/", 3)`.

One issue with implementing that would be how to classify assignment operators: `+=` and `-=` are updateAssign operators, `<=`, `>=`, and `!=` are not.

ghost:
Joe,

I've had a go at implementing user-defined operators using static 
strings, so far it seems to work fairly well. I've updated the variadic 
arithmetic pull request with all the updates. As per your suggestion 
certain characters ( +,-,*,/,\,%,=,<,>) are parsed as strings, reserved 
symbols are filtered out e.g. <--,  and returned as static strings to 
the magic operatorCall function.

An operator can easily be defined using the operator function, e.g.

overload operator(static #"**", BIN) = pow;

  (where BIN is alias of #"binary" flagging a binary (non-comparison) 
operation)

Setting precedence is a little clunky at the moment - uses overloading 
of a multi-valued list of symbols.

Is there any reason why the 'static' keyword is required to make 
#"somestring" static?

If you have any suggestions on how to improve this for possible 
inclusion in Clay let me know.
-Jeremy.

lib-clay/operators/operators.clay
```diff
@@ -330,19 +329,99 @@ inline overload indexUpdateAssign(static F, dest:T, forward ..indicesAndSrc:A) {
...
 
 
+//
+//  operators 
+//
+
+defaultOps() = #"++",#"+",#"-",#"%",#"\\",#"/",#"*";
+precedenceList() = ..defaultOps();
+
+alias BIN = #"binaryOp";
```
jckarter: Instead of aliasing strings, these could be symbols. `define BIN;`

jckarter:
I'm a bit worried about global pollution with this design. In a large program, it would be very easy to end up with conflicting operator definitions, and it would be difficult to figure out what's going on. Maybe operators should be symbols that live in modules, similar to how they work in Haskell. Have `(+)` parse like a name reference and be looked up in the import list, and `a + b * c` parse as `operator(a, mod1.(+), b, mod2.(*), c)` for whatever modules the operators are imported from. Is that crazy?

ghost:
Well, I think "design" is a bit strong, maybe "first pass hack" would be 
more appropriate. Global pollution would certainly be bad. Using symbols 
is probably a sensible idea, although I don't really like the 
parenthesised operator Haskell syntax. Not entirely sure about the C++ 
'operator*' either . . .

ghost:
I've made the change the operator desuger to symbols (my clay/github 
operator branch) which works nicely and is much cleaner than my previous 
"design", however I haven't really had time to delve into the module 
system, is it straight forward to lookup the module name while desugaring?

jckarter:
As long as operator names parse or desugar into NameRef nodes you shouldn't need to modify anything about how module lookup works.

ghost:
Yeah, I see that now, thank you. I'll try to get variadic update 
assignment working correctly before I update the pull request.

ghost:
Updated to allow parenthesized operator symbols in function definitions. Operators are desugered to NameRef and returned to operator module for evaluation. Update assignment now takes variadic operator arguments and returns variadic op list to operator module.

ghost:
Joe, I think this branch is ready. I'll work on sequence operations separately.

I also have a branch (clay/when) with all these updates and the predicate pattern guard <code>|</code> changed to the keyword <code>when</code> if that would be useful. Predicates seem to be more readable with this change.

Let me know if you are intending to pull in the string-integration branch before anything else and I'll hold off doing anything new. I'll also merge string-integration with the branches I have on pull-requests.

jckarter:
The `|` to `when` change sounds good, and frees up `|` as an operator character. I need to bring string-integration up to date with master, then I'll merge it so you can update your branches.

One thing I couldn't discern from looking at your patches: does referring to an operator with a qualified name, as in `operators.(+)`, work?

jckarter:
Also, are you able to define or overload operators with forms like `(+)(a, b) { ... }`, `overload (+)(a, b) { ... }`, or `overload module.(+)(a, b) { ... }` ?

ghost:
Yeah, the parenthesized operator definitions are just normal symbols. 
New operators can be defined by creating a function using the 
parenthesized operator directly or just alias an existing function, as I 
have in the operators module for the standard operators.  Only valid 
"operator" characters are allowed in the parenthesized definitions etc etc.

I'm not entirely happy with the operatorCall reduction algorithms. They 
could do with some improvement to make use of variadic operator 
functions rather than just using binary functions. I'll get around to 
this sometime. Need a few more predicates in there too.

jckarter:
Would it help to add any features to argument pattern matching? Being able to match arguments after a variadic argument, for example `foo(a, ..b, c) {}` or `foo(..a, b)`, could help with left-associativity in particular.

jckarter:
I merged the string-integration branch. You no longer need `static` to match static strings (which are now just string literals). `foo("bar") {}` will just work.

ghost:
I have actually found myself wanting to match arguments after variadics 
args on occasion - I think it would be a useful feature. I noticed 
variadic assignment is on the issues list too, that would be pretty 
useful too.

PR #242, opened 2012-03-25
Add operatorCall overloads for basic binary op calls

ghost:
Basic overloads for binary ops that don't require precedence evaluation, brings test suite performance back to pre-operatorCall.


lib-clay/operators/operators.clay
```diff
@@ -390,6 +390,12 @@ overload operatorCall(forward ..args)
     = forward compareOp(..noPrecBinaryOp(..foldOp(binaryOp, #countValues(
         ..restValues(..precedenceList())), ..args)));
 
+[T,U,Fn when BinaryOpDefined?(Fn,Type(T),Type(U)) or CompareOpDefined?(Fn,Type(T),Type(U))]
+overload operatorCall(#T, #Fn, #U) = forward Fn(T,U);
```
jckarter: Is this separate overload for static operands actually necessary? It should be covered by the following overload.

jckarter:
Compile-time performance is much better with that special case; thanks!

PR #244, opened 2012-03-27
Custom prefix operators + extras

ghost:
Added:
- custom prefix operators
- separate infix/prefix operator primitives
- Symbol?  primitive
- support for non-return operators
- internal operator names no longer use parens

Removed:
- plus/minus primitives (use add/subtract)

Notes:  All operator symbols must be unique (exception is the addressof operator which is defined in the compiler). Tests/libs updated etc.


compiler/src/parser.cpp
```diff
@@ -157,9 +157,9 @@ static bool boolLiteral(ExprPtr &x) {
     return true;
 }
 
-static string cleanNumericSeparator(int op, const string &s) {
+static string cleanNumericSeparator(string op, const string &s) {
```
jckarter: The string arguments should be passed by `const &` since they don't need to be copied.

compiler/src/analyzer.cpp
```diff
@@ -2373,6 +2373,8 @@ MultiPValuePtr analyzePrimOp(PrimOpPtr x, MultiPValuePtr args)
     case PRIM_TypeAlignment :
         return new MultiPValue(new PValue(cSizeTType, true));
 
+    case PRIM_ProcedureP :
```
jckarter: If you don't need the `Procedure?` primitive, I think it should be removed. The fact that procedure symbols are internally different from other symbols is an implementation artifact I'd like to eventually remove.

jckarter:
Looks good. Would it be too much work to replace the old operator names (`add`, `subtract`, etc.) with the operator names instead of having the operators be aliases? That would eliminate the need to have a different `bitxor` operator just to keep unary `bitnot` and binary `bitxor` distinct.

ghost:
OK, just to make sure I follow you on this . . .

define (+);
define (-);
etc

rather than aliasing add, subtract, etc.

Use overloaded (~) function for both bitnot and bitxor.

jckarter:
Yeah. You could also alias the old names if you think that would be useful, `alias add = (+);` etc.

jckarter:
On my machine, the llvm/sse/intrinsics/1 test fails:

```

[llvm/intrinsics/sse/1]
Failure: compiler error
Error:

 main() {
     println(x86_sqrtsd(Vec[Double, 2](4.0)));
 -------------------------------------^
 }

 /Users/joe/Documents/Code/others/clay/test/llvm/intrinsics/sse/1/main.clay(4,37): error: no matching overload found
     85 universal overloads not shown (show with -full-match-errors option)

 compilation context: 
     Vec[Float64, 2](Float64)
   /Users/joe/Documents/Code/others/clay/test/llvm/intrinsics/sse/1/main.clay(4,37):
     main()
   /Users/joe/Documents/Code/others/clay/build/compiler/src/../../../lib-clay/system/system.clay(32,33):
     getExitCode()
   /Users/joe/Documents/Code/others/clay/build/compiler/src/../../../lib-clay/system/system.clay(32,26):
```

ghost:
Yeah i notice that, it needed simd import. I've pushed the fix to the 
prefix branch.

jckarter:
Do you intend to tweak the operator implementation much more after this? If it's ready to settle down it might be time to describe the changes in the ChangeLog and/or a post to the Google group.

PR #248, opened 2012-03-29
'@' for addressof and custom (prefix-/update-)Assign <op>:

ghost:


jckarter:
Ooops, I just merged my iterator-protocol branch, so you might need to go through and update again. Sorry.

ghost:
OK, i've also tried using '<op>:=' for updateAssign, keeping 
reassignment as '='.  it works however i'm not too sure about syntactic 
consistency . . .

jckarter:
Yeah, I'm not a huge fan of `<op>:=`, it was just something that came to mind.

ghost:
Nonetheless it seems to work pretty well with custom operators e.g.

foo(a, b) = a + b;

alias (~~) = foo;

main() {
     var a = 2;
     a ~~:= 2;
     println(a);
}

gives correct output. So just need to decide on syntax.

ghost:
How about a leading/trailing colon and drop the equals altogether?

a :+ 2;

or

a +: 2;

too cryptic?

ghost:
Another idea which looks somewhat cleaner although a little longer is to 
use a keyword e.g.

update a \* 2;

jckarter:
Of those choices `a +: 2;` looks the nicest to me.

ghost:
I've put in the updateAssign stuff.

One thing more that may be nice is prefix update assign, which is now possible with the new unambiguous updateAssign operators e.g. <code> -:a </code> negates a 'in-situ'

What do you think?

ghost:
This branch also has the fix for issue#249

PR #251, opened 2012-03-31
User defined operator precedence/associativity

ghost:
Improved infix operator evaluation - still uses overload pattern matching though which can probably be improved.

Fixed math lib bugs
Removed beta function from lib (temporary)


test/math/native/out.txt
```diff
@@ -1010,7 +1010,7 @@ beta(3,1) = 0.3333333333333333
 beta(3,-1) = nan
 beta(3,3) = 0.03333333333333334
 beta(3,38) = 3.373819163292818e-05
-beta(3,355) = 4.432854008828962e-08
+beta(3,355) = 4.432854008828726e-08
```
jckarter: Why did these results change?

ghost: Looking into that. Odd that its just the beta function which uses 
gamma/exp etc but the results for those functions alone didn't change . .

lib-clay/operators/operators.clay
```diff
@@ -423,56 +368,171 @@ inline overload prefixOperator(fn, forward x)
...
+
+overload boolOp(a) = a;
+
+
+inline overload infixOperator(..args) = boolOp(..binaryOp(..args));
+[OP]
+inline overload infixOperator(a, #OP, b) = OP(a,b);
+
+[OP,T,U when integerEquals?(countValues(..CallOutputTypes(OP,T,U)),0)]
+inline overload infixOperator(a:T, #OP, b:U) {OP(a,b);}
```
jckarter: You can have a single overload  `inline overload infixOperator(forward a, #OP, forward b) = forward ..OP(a,b);` instead of separate ones for no- and single-returns

ghost: Yeah that makes sense i'll update it.

lib-clay/operators/operators.clay
```diff
@@ -423,56 +368,171 @@ inline overload prefixOperator(fn, forward x)
...
-[OP,T,U when BinaryOpDefined?(OP,T,U)]
-inline overload noPrecBinaryOp(forward a:T, #OP,forward b:U, forward ..args)
-    = forward ..noPrecBinaryOp(OP(a,b), ..args);
-[OP,T,U when integerEquals?(countValues(..CallOutputTypes(OP,T,U)),0)]
-inline overload noPrecBinaryOp(forward a:T,#OP,forward b:U, forward ..args) 
-    {OP(a,b);}
-[OP,T,U when BinaryOpDefined?(OP,T,U)]
-inline overload noPrecBinaryOp(forward a:T,#OP,forward b:U)
-    = forward OP(a,b);
+overload binaryOp(..args) = ..args;
```
jckarter: The `binaryOp` overloads should probably all be inline, and forward their arguments and return values

ghost:
Spot of insomnia so i fixed up the math libs. I think thats all issues fixed with the exception of checking IEEE special cases.
I'll replace the beta function when i get time, can't imagine anyone is using it at this point.

lib-clay/operators/operators.clay
ghost: Not convinced by the pipe idea you suggested for handling variadic 
inputs to infix operators. I guess the issue is the evaluation of 
operand expressions prior to passing the infix chain into infixOperator. 
One solution could be to pass through the unmodified section of the AST 
and handle in that in the library. However that seems impractical.

I'm still concerned about compile times and that the 
infix-/prefix-Operator setup is going to balloon into a full blown term 
re-writing thing. Exposing/modifying  AST should probably be done from a 
single point **AST** or whatever.
So I'm wondering whether it would be better to move operator handling 
back into the parser, use top level symbol definitions for custom 
operators and use the afore mentioned Pratt parser to sort it out.  You 
would still get the user defined ops/prec/assoc just tabulate operator 
specs during parsing as well as processing symbol as usual i.e. define 
(**)[LEFT,1000] or define (**, LEFT, 1000) or whatever.

Or do you think the current approach can be sufficiently extended 
without getting messy?

jckarter: On Sun, Apr 1, 2012 at 3:44 PM, agemogolk <reply+p-631168->

>  Not convinced by the pipe idea you suggested for handling variadic
> inputs to infix operators. I guess the issue is the evaluation of
> operand expressions prior to passing the infix chain into infixOperator.
> One solution could be to pass through the unmodified section of the AST
> and handle in that in the library. However that seems impractical.
> 
> I'm still concerned about compile times and that the
> infix-/prefix-Operator setup is going to balloon into a full blown term
> re-writing thing. Exposing/modifying  AST should probably be done from a
> single point **AST** or whatever.

I had been thinking that an **AST** operator and `eval` support for
ASTs is ultimately the right way to go with this. But as you said,
some cute potential uses of variadic arguments are probably not worth
the effort; it's really just an accident of the desugaring that they
happen to work.

> So I'm wondering whether it would be better to move operator handling
> back into the parser, use top level symbol definitions for custom
> operators and use the afore mentioned Pratt parser to sort it out.  You
> would still get the user defined ops/prec/assoc just tabulate operator
> specs during parsing as well as processing symbol as usual i.e. define
> (**)[LEFT,1000] or define (**, LEFT, 1000) or whatever.
> 
> Or do you think the current approach can be sufficiently extended
> without getting messy?

Aside from the variadic arguments issue, are there any other
outstanding issues with your current implementation? It seems like
we're closing on a usable interface based on the current low-level
interface.

-Joe

ghost: Outstanding issues are:
     \* variadic list operands (including multi-valued lists)
     \* compilation performance
     \* seperate handling of infix/prefix operators
     \* repeated operators to variadic calls e.g.  1+2+3+4_5_6 should give  (+)(1,2, (_)(4,5,6)) instead of (+)((+)(1,2), (_)((*)(4,5),6))

Implementing anything that uses variadic lists brings up the same issue I had with mixing evaluation of infix & prefix operators - differentiating static operands from operators. The only thing making the current infix implementation work is (was!) the strict operand-operator-operand form. Making variadic list operands work would require an Operator? primitive which is difficult due to the possible definition of an operator via global aliases which is an issue we've discussed before.
For example using this type of operator definition 'alias (**) = pow;' the infixOperator desuger will actually look like:

2 *\* 3 --> infixOperator( 2, Static[pow], 2) 

where 'pow' obviously can't be treated as an operator symbol.

Although, I think if we have operators defined as symbols we really do need to have operator symbol introspection. Disallowing global alias definitions of operators is a solution, but not a good one as it would introduce an inconsistency in the 'operator is just a symbol' paradigm and 'alias (**) = pow;' is pretty useful.

That restriction aside the simple solution to variadic list operands is to disallow them. However, it would be nice to be able to use multi-valued lists as operands.

An alternative is to pass them through with some kind of delimiter that can processed unambiguously, which I guess is the 'pipe' idea e.g.

infixOperator( 1,  (+), <var_list> , 2, 3, </var_list> , (_),4)   -->  (+)(1, (_)( 2, 3, 4))

Probably the best solution is to find a way of making Operator? possible.

jckarter: On Mon, Apr 2, 2012 at 6:57 AM, agemogolk
reply@reply.github.com
wrote:

> I think the best solution is to find a way of making Operator? possible.

How about passing the operators and operator positions out-of-line as
symbol parameters, so that `x * (y,z) + w` becomes
`infixOperator[(*),1,(+),3](x,y,z,w)`? That allows prefix and infix
operators to be passed together in a recoverable way, and I think the
code to decompose that form would be simpler and thus faster as well.

-Joe

ghost: On 02/04/12 16:20, Joe Groff wrote:

> > -private define noPrecBinaryOp;
> > -[OP]
> > -inline overload noPrecBinaryOp(forward a, #OP,forward ..args)
> > -    = forward a,OP,..noPrecBinaryOp(..args);
> >   -[OP,T,U when BinaryOpDefined?(OP,T,U)]
> >   -inline overload noPrecBinaryOp(forward a:T, #OP,forward b:U, forward ..args)
> > -    = forward ..noPrecBinaryOp(OP(a,b), ..args);
> >   -[OP,T,U when integerEquals?(countValues(..CallOutputTypes(OP,T,U)),0)]
> >   -inline overload noPrecBinaryOp(forward a:T,#OP,forward b:U, forward ..args)
> > -    {OP(a,b);}
> >   -[OP,T,U when BinaryOpDefined?(OP,T,U)]
> >   -inline overload noPrecBinaryOp(forward a:T,#OP,forward b:U)
> > -    = forward OP(a,b);
> >   +overload binaryOp(..args) = ..args;
> >   On Mon, Apr 2, 2012 at 6:57 AM, agemogolk
> >   reply@reply.github.com
> >   wrote:
> >   I think the best solution is to find a way of making Operator? possible.
> >   How about passing the operators and operator positions out-of-line as
> >   symbol parameters, so that `x * (y,z) + w` becomes
> >   `infixOperator[(*),1,(+),3](x,y,z,w)`? That allows prefix and infix
> >   operators to be passed together in a recoverable way, and I think the
> >   code to decompose that form would be simpler and thus faster as well.
> 
> -Joe

I've been looking at the out-of-line with direct recursion (rather than 
forcing it through the overload system) possibility on the side, but 
hadn't yet considered passing positions - looks like it could work. 
Would still be nice to have Operator? though.

PR #252, opened 2012-04-01
Fix building on Fedora

solson:
I'm helping @CodeBlock make a Clay package for Fedora, and he ran into this problem from `cmake`:

```
CMake Error at compiler/src/CMakeLists.txt:69 (set_target_properties):
  set_target_properties called with incorrect number of arguments.
```

This commit fixes it.


jckarter:
Looks good; thanks @tsion and @CodeBlock!

PR #253, opened 2012-04-02
lgamma bug fix + add beta function + remove redundant operator code

ghost:


jckarter:
It looks like you didn't add back the math.native.beta module.

PR #256, opened 2012-04-06
Fix math lib bugs + update tests

ghost:
Check special case results (nan/inf)
Fix casting bugs
Add more test cases


lib-clay/math/native/native.clay
```diff
@@ -56,12 +56,12 @@ overload recip(x:T) = T(1) / x;
 inline overload abs(x:I) = if(x >= 0) x else -x;
 
 inline overload sign(x:Long) = bitor(bitshr(x,63),Long(bitshr(ULong(-x),63)));
-inline overload sign(x:Int) = bitor(bitshr(x,31),Int(bitshr(UInt(-x),31)));
+inline overload sign(x:Int) = bitor(bitshr(x,31),wrapCast(Int32,bitshr(UwrapCast(Int32,-x),31)));
```
jckarter: I think you mean `wrapCast(UInt32,...)` here. Is there a test for `sign(Int)`?

lib-clay/math/native/native.clay
```diff
@@ -56,12 +56,12 @@ overload recip(x:T) = T(1) / x;
...
 inline overload sign(x:Short) = bitor(bitshr(x,15),Short(bitshr(UShort(-x),15)));
 inline overload sign(x:Byte) = bitor(bitshr(x,7),Byte(bitshr(UByte(-x),7)));
 
 [T when Float?(T)]
-overload significand(x:T) = scalb(x,T(-ilogb(x)));
+overload significand(x:T) = scalb(x,T(wrapSubtract(ilogb(x))));
```
jckarter, line 64: `wrapSubtract` for one argument is defined incorrectly at numbers/numbers.clay:526. Is `significand` tested?

ghost, line 64: The correction for wrapSubtract(a) is in this pull request.

Significand is in the native test suite already.

lib-clay/math/native/native.clay
ghost: Thanks for spotting those. I still need add a few math tests.
There are also bugs in the range reduction of sin,cos,tan for large 
arguments, which is my current priority.

PR #260, opened 2012-04-09
Add float32 math lib tests

ghost:
Fix exp(Float32) function
Update math test/output


jckarter:
On macosx, `scalbf`, `scalbl`, `significandf`, and `significandl` are missing. The man page for `scalb` says that it's deprecated and the `scalbn{f,l,}` functions should be used instead. Could the math module always use the native `significand` implementation, and `scalbn` in favor of `scalb`?

PR #261, opened 2012-04-10
Decouple some modules from prelude

ghost:
Modules removed from prelude:
algorithms
algorithms.strings
hash
hashmaps
uniquepointers
sharedpointers
referencetypes

Modules removed:
monads (trivial content moved to operators)

Imports updated
Tests updated


jckarter:
It's a good start, thanks. I think ideally the prelude should only include the basics to let you program C-with-value-semantics, so things like math, vectors, io, printer, and strings (not including strings.literals) should probably be pushed out as well.

PR #267, opened 2012-04-11
Update prelude/core

ghost:
Remove io.files and io.streams from prelude.
Remove lambdas, complex ,strings, cstringrefs from prelude.
Remove numbers/parser from the core imports, removes all hooks to printer.
Move some functions from floats to math/native.
Also add cbrt overloads

Move core modules to core. Import core into prelude.
libc functions used in the core modules reside in core.libc, all other modules use the genned libc


ghost:
Joe, I've restructured the library to use a core set of modules as discussed. I've removed what I can from the core set, but vectors etc are still there. It may be possible to remove vectors, but it would require some changes that may not be worth it e.g. system module uses Vector[Char] to handle commandline args.

The printer/io system is properly decoupled from the core now. Running clay without printer stuff cuts load time pretty much in half. Compile times are reduced significantly overall,. for example the test suite takes 130-135 secs for me which is a huge drop from the 260-400 sec from a couple of weeks ago.

jckarter:
I don't like `libc` living in core with a magic alias in everyone's namespace. How hard would it be to pluck out only the libc definitions used in core into `core.libc`, with the full generated bindings remaining in `libc`?

The system module could use `CoordinateRange[Pointer[Char]]` or `CStringRef` to reference commandline arguments, since the process owns the argv memory. Are there other parts of core that need vectors?

ghost:
No problem taking out the core.libc stuff. It's only malloc, strlen, abort, etc.

The errors module also uses Vectors.
Sequences can also be removed if vectors go. The only small issue there is a couple of functions in arrays using zipped which should be easy enough to get around.

ghost:
I've move the full libc bindings out of core and created core.libc modules for each platform. These aren't generated automatically and will need testing.

jckarter:
The lexer patch I just pushed gives the test suite a pretty decent speedup on my machine—290s down from 330s. How's the timing with your library cleanup on your machine?

ghost:
~135s down from ~260s

ghost:
I merged your lexer patch and it there's no significant speed up for me, 
however that's with all the lib changes too.

lib-clay/expat/generated/Makefile
```diff
@@ -1,2 +1,2 @@
 generated.clay: 
-	clay-bindgen -match expat -import unix -import libc -o generated.clay generated_from.h
+	clay-bindgen -match expat -import unix -import core.libc -o generated.clay generated_from.h
```
jckarter: I think this should be changed back to `libc`

lib-clay/llvm/libclang/generated/Makefile
```diff
@@ -1,5 +1,5 @@
 generated.clay: generated_from.h
 	clay-bindgen -o generated.clay \
-	    -import libc \
+	    -import core.libc \
```
jckarter: This too

jckarter:
Is it actually much of a win to separate `libc` and `math.libm`? It looks like `<math.h>` pulls in most of libc, and I guess there's no good portable way to filter it down to only the math.h definitions.

ghost:
<math.h> doesn't pull in much of libc at all on my system.  The 
generated linux64 libm file just has the math functions and a couple of 
aliases. I made the comparison of with/without split prior to 
introducing the core.libc.

While on the subject of bindings is it possible to suppress the import 
of functions duplicated with leading underscores?

jckarter:
I see, you just copied the libc bindings for non-linux platforms. In libc.windows.{32,64}.clay there are some shims because some libm functions are implemented as macros; those will need to be moved to a `math.libm.platform` module.

ghost:
I'm just testing whether seperete libm makes a difference to performance with all the other changes in place.

jckarter:
If it's not a big difference I would vote for keeping libc and libm in one module for simplicity's sake; fewer binding modules to maintain.

ghost:
Sure, that would be preferable.

Ok, so ~3s slower running the test suite with monolithic libc/m 
bindings. Given the variability in timing due to system stuff I guess 
theres no need to split it up.

ghost:
OK. I think thats about it for my tinkering. 

One niggle though, printer imports numbers.floats just to get nan/nan?.
I've tried to break printer down a bit to reduce overhead, maybe this could be taken further e.g. requiring printRepr to imported explicitly . .

jckarter:
`nan` and `nan?` should probably live in core somewhere. `printRepr` and the `repr` wrapper from `printer.formatter` could probably go in their own module.

ghost:
Joe,

Did you find much of a performance gain with the core lib changes?

jckarter:
I didn't notice much change in test run time. It may be that the loader is no longer a significant part of compile time after both the lexer and core reorganization changes; we should probably look at profiles again to see where time is being spent now.

ghost:
Hmm, interesting. I wonder if I missed something in mac modules that 
still hooks into printer/io stuff. Still haven't got llvm/clang/clay up 
and running on the macbook, it was a pain last time i looked, but I 
guess this is motivation to do so.

I'll take another look at the load time breakdown. A quick valgrind did 
show ~300,000 fewer heap allocations for a trivial compilation. Still 
leaks though (~23 meg), although the cause may be the llvm passes - 
peepholeoptimizer showed up.

PR #280, opened 2012-05-02
Fix for nthValue() in lib-clay/values/values.clay

trss:
Fixed the bug nthValue() not returning lvalue when needed, by adding a couple of missing 'forward's.


jckarter:
Looks good, thanks!

PR #286, opened 2012-06-11
adding claydoc.

aep:


jckarter:
@aep I noticed that `make claydoc` rebuilds everything redundantly, even though the same source files are used for the compiler. It might make sense to move the claydoc cpp files and cmake rules into compiler/src, so it's easier to cajole cmake into reusing the same object files for both.

aep:
yeah, but still recompiles them for me, no matter if i put them in the same dir.

aep:
@jckarter ^

PR #292, opened 2012-07-21
claydoc fixes

aep:


aep:
there we go. thanks for making me do that, spotted more bugs on the way. ;)

PR #293, opened 2012-07-21
claydoc fixes

aep:
last one for today, really ;)


jckarter:
Merged. You might also tweak your test case so it exercises the fix.

Issue #266, opened 2012-04-11
'import foo; import foo.bar;' gives improper name redefinition error


jckarter:
This should work but doesn't:

```
import printer;
import printer.formatter;
```

jckarter:
Fixed by @agemogolk in #299

PR #299, opened 2012-07-31
Fix module import issue #266

ghost:
Fixed import name redefinition error.
Add test for issue#266 fix.


jckarter:
Looks good @agemogolk; thanks!

PR #303, opened 2012-08-04
rewrite values test using clay test framework

stepancheg:
this makes test easier to read


jckarter:
Looks good @stepancheg, thanks!

PR #305, opened 2012-08-04
filterValues function

stepancheg:
unlike mapValue, filterValues parameters must be static values


jckarter:
Looks good, thanks!

Issue #35, opened 2011-12-10
Multiple-value variable bindings


jckarter:
It should be possible to bind multiple-value variables.

```
var zero, one, ..rest = integers(static 10);
```

jckarter:
Fixed by @agemogolk

Issue #246, opened 2012-03-28
Allow variable bindings to declare types, bind pattern variables


jckarter:
`var` etc. bindings should be able to declare types, e.g.:

```
var x:Int = 42;
```

It would also be convenient sometimes to be able to bind pattern variables from variable types, to extract otherwise difficult-to-derive type values:

```
var [E] elt:E = seq[0];
```

ghost:
Joe, I've been tinkering with updates to the local bindings and managed to get them mostly working with the pattern matching system, which allows variadic assignment, pattern variables, typespecs, etc. The code is currently a little hackish and requires some filling out for error reporting and such. I've pushed it to my 'bindings' branch if you want a look (changes are mostly in codegenBindings, analyzeBindings, and the parser). Do you think this is a suitable solution or did you have something else in mind?

The current implementation breaks implicit unpacking, which can no doubt be fixed up. However I wonder if this should be kept in the language as it feels inconsistent with the use of ellipsis for all other multi-value unpack/variable statements.  This is particularly so when assigning to multi-value variables e.g. <code>var a, ..b = implict();</code>. I feel this should really be explicit and look similar to assigning a variadic variable like <code>var a, ..b = ..c; ..b = ..d; var a, ..b = ..explicit();</code> and so on.

What do think about making all multi-value unpacking explicit?

jckarter:
Excellent. Re: implicit unpacking, It used to be that `..` was always required to be explicit, but that requirement was a common complaint from users. The unpack does feel redundant in cases like `var ..a = multiValues();`—the left-hand side of the binding makes it clear that multiple values are expected. If you could retain that behavior, that would be ideal. Currently it handles implicit unpacking by passing a `wantCount` greater than 1 to functions such as `codegenMulti` and `analyzeMulti`; a hacky but effective way to accomplish this for variadic bindings would be to just add 2 to the wantCount for a variadic binding.

jckarter:
Is your math module pull request ready to be merged? There are a bunch of extraneous diffs in the master...bindings comparison diff.

ghost:
Yeah, I think the math module stuff can be merged, however there may be 
an issue with non-linux test output. Probably just need to regenerate 
stuff. All tests pass for me. I've checked linux output and everything 
works as it should with regard to IEEE special cases. Just a few epsilon 
differences between native and libm output. The tests have been updated 
to use codepointers loops rather than deep nested static_for loops. This 
makes the math tests considerably faster.

[...]

PR #322, opened 2012-08-10
Update local bindings (multiple value, typespec, patterns) fixes issues #35 & #246

ghost:
Local bindings update
- multiple value bindings
- typespecs
- patterns/predicates in local scope
- var/ref/forward semantics preserved
- alias binding not modified

Added simple test


test/uncategorized/16/main.clay
```diff
@@ -7,7 +7,7 @@ enum Kind (
 );
 
 main() {
-    var a, b, = UGLY, BAD, ;
+    var a, b = UGLY, BAD, ;
```
jckarter: Why was the comma removed? Does the trailing comma not work anymore?

test/bindings/variadic/main.clay
```diff
@@ -0,0 +1,21 @@
...
+	
+	var [..I] a:Int32, ..c:I = 4, 5, 4, 5, 6, 7;
+	var ..b = ..foo(..c);
+	ref ..j = ..b;
+	println(..j);
+	println(..c);
+
+	var zero, one, ..rest = integers(#10);
+	println(..weaveValues(", ", ..rest));
+}
```
jckarter, line 21: Maybe also add a `println(..weaveValues(", ", ..I));` to show that `I` binds to the right values `Int, Int, ...`

ghost:
I've allowed a trailing comma in the binding list. Eventually i'll add some more tests for the bindings.

PR #331, opened 2012-08-13
io.errors, os.errors: add failed function name to exception

stepancheg:


stepancheg:
Included in https://github.com/jckarter/clay/pull/333

PR #335, opened 2012-08-14
launcher: do not check for directory when -run option is specified

stepancheg:


stepancheg:
Updated patch: do not remove output on signal.

Issue #239, opened 2012-03-20
Argument/value matching after variadics


jckarter:
It would be useful to be able to bind single arguments or variables after a variadic, for example:

```
foo(a, ..b, c) {
    var d, ..e, f, g = a, ..b, c;
}
```

jckarter:
Done by @agemogolk in #336.

PR #336, opened 2012-08-15
Match after variadic - fixes issue #239 & #312

ghost:
Can now match argument lists and bindings using variadic variable in any position, see issue #239
Added simple test.
All tests pass.
Small change to values module to use new feature.
Note the weirdness of alias functions has been preserved . . .

Have also added some better error messages for staticToInt & staticToBool - fixes issue #312


jckarter:
Looks good. What's the weirdness with alias functions? (Alias functions are weird in general, of course.)

ghost:
No changes with the way alias functions work, just probably seem a little weirder with the matching after variadics. The way expressions are passed may be confusing to some.

jckarter:
Yeah, it's definitely weird, but I can't think of a good way of breaking apart variadic alias arguments without evaluating them first. The compiler should perhaps be less permissive with alias variadic arguments and raise an error if it can't match expressions to arguments in a way that makes semantic sense. IIRC currently `alias foo(x)` called like `foo(..y)` will try to bind `x` to `..y`.

ghost:
Yeah thats right, it's pretty straight forward to do a check during codegen if we can decide on the limitations.

ghost:
Would a one-to-one and multi-to-multi mapping be too limiting? Maybe could use a single multi-value as the alias function argument to capture all expressions in the call as a special case . . . ?

jckarter:
You can already capture all expressions using `alias foo(..all)`. I don't think there's any code that intentionally takes advantage of the ability to bind single symbol to multiple-value expressions.

ghost:
OK, so I'll add checks to stop the compiler trying to put a multi-values into single expressions. Also, should this be allowed:
<code> alias foo(..a, b, ..c, d) </code> or just stick to a single variadic?

jckarter:
I don't think that should be allowed, since it isn't allowed for normal functions.

Issue #103, opened 2011-12-31
Strong typedefs


jckarter:
There should be a feature like Haskell's `newtype` that creates a type that's representationally identical to another, but has a distinct type.

ghost:
Would a new keyword be necessary? Maybe reuse define e.g.
<code>define MyInt = int32;</code>

<code>[T,U when Integer?(T,U)] //overide predicate of original
define MyRecord[T,U] = OriginalRecord[T,U];</code>

or would that be too confusing?

What about printed name formatting in compiler output - maybe something like: 
<code>MyInt<Int32></code>

jckarter:
Overloading `define` doesn't seem like a good idea. My plan was to have something along the lines of `newtype A = B;`.
I think the printed name should be formatted normally, the same as any other symbol.

ghost:
Strong typedef implementation "newtype" added in #388

PR #364, opened 2012-08-26
Implement newtype (strong typedefs) - fixes issue #103

ghost:


compiler/src/loader.cpp
```diff
@@ -235,6 +236,22 @@ static void installGlobals(ModulePtr m) {
             }
             break;
         }
+        case NEW_TYPE : {
+            NewType *nt = (NewType *)x;
+            PatternPtr pattern = evaluateOnePattern(nt->target, m->env);
```
jckarter, line 241: The `newtype` target could be evaluated as an expression (like variant or record bodies) instead of being restricted to a pattern.

compiler/src/clone.cpp
```diff
@@ -482,4 +482,151 @@ void clone(const vector<CatchPtr> &x, vector<CatchPtr> &out)
...
+
+    default :
+        assert(false);
+
+    }
+
+    return out;
+}
+
+}
```
jckarter, line 632: What is all this new code for?

compiler/src/loader.cpp
```diff
@@ -235,6 +236,22 @@ static void installGlobals(ModulePtr m) {
...
+            ObjectPtr obj = derefDeep(pattern);
+            if (obj == NULL) {
+                error(nt->target, "target type undefined");
+            } else if (obj->objKind != TYPE){
+                error(nt->target, "not a valid type");
+            } else {
+                Type *t = (Type *)obj.ptr();
+                TypePtr newObj = clone(t);
```
jckarter, line 249: Rather than clone the type object, you could handle newtypes in types.cpp by generating the same llvm::Type for the newtype as for the target type. Look at how variants are implemented in terms of their variantRepr type.

Issue #123, opened 2012-01-02
Reduce excessive allocations in parser


jckarter:
Since AST nodes aren't ever really freed once created, they should be allocated using a bump allocator and not refcounted or freed.

jckarter:
Done by @agemogolk in #365

PR #365, opened 2012-09-01
Use BumpPtrAllocator for Object allocation

ghost:
Significantly reduces number of allocations during parsing and compilation.

This is pretty simple implementation designed to minimise source changes, but seems to work ok. 
Maybe addresses issue #123.

### DIFF
--- a/compiler/src/clay.hpp
+++ b/compiler/src/clay.hpp
@@ -332,8 +332,6 @@ public :
     }
 };
 
-static llvm::BumpPtrAllocator *ObjectAllocator = new llvm::BumpPtrAllocator();
-
 
 //
 // Object
@@ -349,12 +347,6 @@ struct Object {
     void decRef() {
         if (--refCount == 0)
             delete this;
-    }
-    void *operator new(size_t num_bytes) {
-        return ObjectAllocator->Allocate(num_bytes, llvm::AlignOf<Object>::Alignment);
-    }
-    void operator delete(void *p) {
-        ObjectAllocator->Deallocate(p);
     }
 };
 
@@ -1007,10 +999,18 @@ void tokenize(SourcePtr source, size_t offset, size_t length,
 // AST
 //
 
+static llvm::BumpPtrAllocator *ANodeAllocator = new llvm::BumpPtrAllocator();
+
 struct ANode : public Object {
     Location location;
     ANode(int objKind)
         : Object(objKind) {}
+    void *operator new(size_t num_bytes) {
+        return ANodeAllocator->Allocate(num_bytes, llvm::AlignOf<ANode>::Alignment);
+    }
+    void operator delete(void *p) {
+        ANodeAllocator->Deallocate(p);
+    }
 };
 
 struct Identifier : public ANode {
--- a/compiler/src/clay.hpp
+++ b/compiler/src/clay.hpp
@@ -317,7 +317,7 @@ public :
     }
     T &operator*() { return *p; }
     const T &operator*() const { return *p; }
-    T *operator->() { return p; }
+    T *operator->() { return p; } 
     const T *operator->() const { return p; }
     T *ptr() const { return p; }
     bool operator!() const { return p == 0; }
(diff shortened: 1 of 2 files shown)


jckarter:
Is there any compile-time performance improvement with this patch? How does memory usage compare?

compiler/src/clay.hpp
```diff
@@ -347,7 +348,10 @@ struct Object {
     void incRef() { ++refCount; }
     void decRef() {
         if (--refCount == 0)
-            delete this;
+            ObjectAllocator->Deallocate(this);
```
jckarter: You should call `ObjectAllocator->Deallocate` from `Object::operator delete`, so that the destructor still gets invoked.

ghost:
Significant parse/compile time improvements when compiling individual projects. Test suite time dropped ~20 secs ( ~8%). Allocations seem to be halved at worst for small modules.  The default slab allocation is fairly small (4096 bytes) so memory usage is similar. Larger slabs made negligible difference as most allocations are around 100 bytes.

Be interesting to see how well this works on other platforms.

jckarter:
Well, BumpPtrAllocator never actually releases memory, so I think compile-time-computation-heavy code that allocates lots of temporaries will use more memory or run out of memory. For instance, one of the cocoa.\* tests on OS X tops out at 1.89GB with your patch. While it's sensible to leak AST nodes, invoke tables, overloads, and other objects that are required for the lifetime of a compile job, I think there needs to be a smarter allocation strategy for EValues and CValues used for evaluation and codegen. Maybe a stack of BumpPtrAllocators mirroring the callstack could be made to work, though there will probably be problems with passing objects up the stack.

jckarter:
OTOH, the reference counting for BumpPtr allocated objects is unnecessary (since the object memory is never reclaimed anyway), and removing it would probably give even better performance improvements.

ghost:
Joe, can you try your OSX test with the bump alloc restricted to ANodes. It still give a significant drop in allocations and increased performance.

ghost:
I did try this with the ref counting disabled and it didn't have much impact. Using the bump for just ANodes makes this a moot point anyway as the other Objects need it.

compiler/src/clay.hpp
```diff
@@ -1000,10 +999,18 @@ void tokenize(SourcePtr source, size_t offset, size_t length,
...
 struct ANode : public Object {
     Location location;
     ANode(int objKind)
         : Object(objKind) {}
+    void *operator new(size_t num_bytes) {
+        return ANodeAllocator->Allocate(num_bytes, llvm::AlignOf<ANode>::Alignment);
+    }
+    void operator delete(void *p) {
+        ANodeAllocator->Deallocate(p);
+    }
```
jckarter: A problem with this is that, since `delete` is not dynamically dispatched and all `Object` subtypes are `delete`d by `ObjectPtr`, it will try to delete `ANode`s using `Object::operator delete`, which will try to `free` memory managed by the `BumpPtrAllocator` and crash. Luckily AST nodes don't ever reach zero refcount currently, but if they did, stuff would break.

To make subclass-specific deallocation work in C++ you need to do something like this to make deallocation happen as part of destruction (and then never stack-allocate):

``` c++

struct Base {
    // overload operator delete to do nothing so that `delete base;` just invokes the dtor
    void operator delete(void*) {}

    // invoke the real virtual deallocator as part of the dtor
    virtual void dealloc() { ::operator delete(this); }
    virtual ~Base() { dealloc(); }
};

struct Derived : Base {
    void *operator new(size_t sz) { return CustomAllocator::allocate(sz); }
    virtual void dealloc() { CustomAllocator::deallocate(this); }
};
```

ghost:
Oh, the joy of learning C++!  OK, i'll see if I can make the code look a little uglier and less intuitive, as the designers obviously intended.

jckarter:
Running the tests with only ANodes bump allocated works with about the same memory footprint as before, although it's a bit slower—bump-allocating everything takes 713s, while bump-allocating only ANodes takes 801s. (That's with LLVM compiled with Debug+Asserts.)

jckarter:
And of course, someone on a Mac that doesn't have 8GB of RAM would get a much slower count for the former case.

ghost:
Did you see any increase in performance with this? Do the cocoa tests still require huge amounts of memory?

jckarter:
Memory usage is the same as before. There's a small speed up running the tests.

PR #366, opened 2012-09-02
Disable ref counting and dealloc for ANodes

ghost:
As bump allocator never releases and ANodes persist during compilation no need to ref count and deallocate.
This seems to give another small performance increase.

### DIFF
--- a/compiler/src/clay.hpp
+++ b/compiler/src/clay.hpp
@@ -342,8 +342,8 @@ struct Object {
     int objKind;
     Object(int objKind)
         : refCount(0), objKind(objKind) {}
-    void incRef() { ++refCount; }
-    void decRef() {
+    virtual void incRef() { ++refCount; }
+    virtual void decRef() {
         if (--refCount == 0)
             dealloc();
     }
@@ -1007,6 +1007,8 @@ struct ANode : public Object {
     Location location;
     ANode(int objKind)
         : Object(objKind) {}
+    virtual void incRef() {}
+    virtual void decRef() {}
     void *operator new(size_t num_bytes) {
         return ANodeAllocator->Allocate(num_bytes, llvm::AlignOf<ANode>::Alignment);
     }


compiler/src/clay.hpp
```diff
@@ -1007,6 +1007,8 @@ struct ANode : public Object {
     Location location;
     ANode(int objKind)
         : Object(objKind) {}
+    virtual void incRef() {}
```
jckarter, line 1010: What's the performance effect of this? My gut feeling is that a virtual function call is more expensive than an inc instruction.

ghost, line 1010: Yeah, something has gone awry with my timings, scratch this one.

jckarter:
I don't think there is much improvement left to be had in tweaking the object model as-is. A deeper redesign, where more objects used unique ownership semantics or value semantics where appropriate instead of blanket reference-counted reference semantics, would be the way to go, but is a lot of work.

ghost:
Sure. I was looking at this  due to the recent increase in number of tests, some of which take a long time. Makes the testing cycle a pain, it was just about ok when the test suite ran in 60-100s.

May be worth reorganising the library structure e.g. lib-core, lib-std, lib-contrib, with corresponding test targets. Fast testing of compiler changes would be nice.

jckarter:
Well @stepancheg has been changing tests to use the `test` module, which is a bit slow to compile. Improving the compile time performance of the test module would be nice.

jckarter:
One thing that might help test suite performance is to make the test runner's use of `commandline.options` and text coloring optional. runtests.py doesn't run the test programs with any flags and writes their output to a plain text log, so having a `-Dtest.minimal` flag that disabled those things would save some compile time.

ghost:
Yeah, the overhead on the test module is rather high for most cases. A trivial 'expectEquals' example takes ~3s just to compile.

PR #367, opened 2012-09-04
Make variadic bindings optional + fixes

ghost:
Add missing arity checks in bindings evaluator
Update arity error messages for bindings/assignments
Remove redundant arityMismatchError function
Update tests

### DIFF
--- a/test/multivalues/assignmismatch/2/compilererr.txt
+++ b/test/multivalues/assignmismatch/2/compilererr.txt
@@ -1,1 +1,1 @@
-error: expected 1 value, but received 2 values
+left side has 1 value, but right side has 2 values
--- a/test/multivalues/assignmismatch/2/compilererr.txt
+++ b/test/multivalues/assignmismatch/2/compilererr.txt
@@ -1,1 +1,1 @@
-left side has 1 value, but right side has 2 values
+error: expected 1 value, but received 2 values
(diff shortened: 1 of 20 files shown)


test/runtests.py
```diff
@@ -389,11 +389,13 @@ def runTests(opt):
...
         print "[Tests]"
         for test in testcases:
+            startTime = time.time()
             res, log = results.next()
+            endTime = time.time()
             if res != "ok" and res != "disabled":
                 logfile.write(log)
             logfile.flush()
-            print "%s: %s" % (test.name(), res)
+            print "%s in %fs: %s" % (test.name(), (endTime - startTime), res)
```
jckarter: Can you move the timing after the `:` so the output is still parseable by python's `config` module? Something like `<name>: <status> (<time>)` perhaps

test/multivalues/assignmismatch/2/compilererr.txt
```diff
@@ -1 +1 @@
-left side has 1 value, but right side has 2 values
+error: expected 1 value, but received 2 values
```
jckarter: This seems like a step backward in error message quality.

test/runtests.py
ghost: Oh, meant to delete the timing stuff. Is it worth leaving in?

jckarter: Seems useful to me.

PR #370, opened 2012-09-05
Tests update

ghost:
Reduce lambda passing in test module.
Inline some functions in test module.
Merge array tests.
Update many test module based tests.
Inline some functions in crypto module.

Test suite runs significantly faster (~20s for me)


compiler/src/codegen.cpp
```diff
@@ -3611,7 +3611,7 @@ bool codegenStatement(StatementPtr stmt,
         MultiPValuePtr mpvLeft = safeAnalyzeMulti(x->left, env, 0);
         MultiPValuePtr mpvRight = safeAnalyzeMulti(x->right, env, mpvLeft->size());
         if (mpvLeft->size() != mpvRight->size())
-            arityMismatchError(mpvLeft->size(), mpvRight->size());
+            arityMismatchError(mpvLeft->size(), mpvRight->size(), false);
```
jckarter: Could you add a `/*hasVarArg=*/` comment to the Boolean arguments?

ghost: Done in this branch and the bindings branch in my other pull request.

jckarter:
Just to double-check—the patches in your `bindings` branch are also in `tests`?

ghost:
Yes, not good practice, I know, but i'd forgotten when I made the new 
branch.

jckarter:
No worries, I can merge both. It all looks good.

jckarter:
I see a significant speedup too—327 seconds vs. 410 seconds before. Nice work!

ghost:
There was plenty of low-hanging fruit in the test module. Probably not 
worth doing much more with it until we have a decent in-module test 
framework.

Issue #230, opened 2012-02-29
Pathological memory consumption by deep nested ..for loops


jckarter:
There's no reason for this sort of thing to eat gigabytes of ram, but it does:

https://github.com/agemogolk/clay/blob/d6c27e3df03840686d7b52d1a24daf521d1e5c2a/test/math/libm/main.clay

ghost:
Could shut this one down too and open a ticket for tuning optimisation passes.

PR #372, opened 2012-09-06
Manually add module opt passes - fixes issue #230

ghost:
Disable buggy value propagation pass (see LLVM bug #12503). This is a cause of poor optimisation times and heavy memory consumption in nested static for loops.

Takes a chunk of time of test suite (~23s for me)

Also, some small native/sqrt fixes.

### DIFF
--- a/compiler/src/clay.hpp
+++ b/compiler/src/clay.hpp
@@ -32,6 +32,7 @@
 #include <llvm/ADT/Triple.h>
 #include <llvm/Analysis/DebugInfo.h>
 #include <llvm/Analysis/DIBuilder.h>
+#include <llvm/Analysis/Passes.h>
 #include <llvm/Assembly/Writer.h>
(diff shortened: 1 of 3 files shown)


jckarter:
Nice catch. I wonder what effect disabling value propagation has on runtime performance.

ghost:
I've only seen increases in performance with this change so far e.g. quicksort1 test runs ~0.3s quicker with exceptions/assertions enabled and no quicker without. The mandelbrot shootout example runs in the same time with exceptions/assertions enabled, but ~0.4s quicker without.

jckarter:
Interesting. I wonder if that's because of increased code size due to the optimization. Perhaps the compiler could do its own value propagation in places where it's obvious, such as when referencing integer constants. The unoptimized LLVM the compiler currently generates is extremely naive.

PR #373, opened 2012-09-08
Fix bug in pattern match after variadic

ghost:


jckarter:
What was the bug? Can you add a test case for it?

ghost:
The bug was really an oversight in my implementation - not incrementing 
an 'argsKey' index counter to the correct position when matching a typed 
arg after an un-typed variadic arg. The patch fixes this for both 
argument patterns and bindings.

I'll add more thorough testing for the pattern matching shortly.

ghost:
I've added a simple test case for this bug.

jckarter:
Thanks. Even if the case is indirectly covered by other code it's good practice to have regression tests specifically for the failing case.

PR #385, opened 2012-09-23
Add linker support to -run mode

galchinsky:
Used Cling's <a href="http://root.cern.ch/svn/root/trunk/interpreter/cling/lib/Interpreter/Interpreter.cpp">tryLinker</a> code as an example.
https://github.com/jckarter/clay/issues/300 passes now (it is 'isNative' condition), but i didn't test bitcode libraries


jckarter:
Looks good, thanks. I just tried it here and it works. Does llvm::Linker also support OS X frameworks so that `-framework` can be supported on Mac and iOS as well?

PR #388, opened 2012-09-27
Strong typedefs 

ghost:
Second attempt at implementing "newtype"!

If this looks to be a reasonable approach I'll finish it up and add tests.
Any more useful primitives required?


jckarter:
This looks great. A useful library utility would be `[T] baseCast(x:T) = bitcast(baseType(T), x);`; there should also be a test that that works as expected. Other than that, nice work.

stepancheg:
Tests explaining what newtype really does, would help a lot. For example,

```
newtype IntAlias = Int;

foo(a: IntAlias) {}

main() {
    foo(1);
}
```

I have no idea whether this works or not.

Or vice versa, when IntAlias is passed where Int expected.

Or this:

```
record Foo ();

overload Foo(a: Int) = Foo();

newtype Bar = Foo;

main() {
    Bar(1)
}
```

and so on.

ghost:
@stepancheg neither of your examples above would compile as the newly created types are just that - new types. `Bar` is not `Foo`, it just shares the same underlying structure. As we don't have something like Haskell's "deriving" the infrastructure to make the type useful has to be added manually.

jckarter:
That's a good idea, @stepancheg. A newtype is like a record with a single field or a variant with a single member, but guaranteed to share layout with the single contained type. As with all Clay types, newtypes are not implicitly castable to their contained type, and so can have their own set of operations independent of the base type. This can be also be used for implementation hiding, for example:

```
private record PointImpl (x:Float, y:Float);
newtype Point = PointImpl;

x(p:Point) = bitcast(PointImpl, p).x;
y(p:Point) = bitcast(PointImpl, p).y;
```

The library should set up newtypes so that they inherit the value semantics of their base type by default.

PR #392, opened 2012-09-30
Implement final overloads

ghost:


compiler/src/analyzer.cpp
```diff
@@ -1832,6 +1832,14 @@ InvokeEntry* analyzeCallable(ObjectPtr x,
         matchFailureError(failures);
     }
 
+    if (x->finalCheck) {//Check final overload
+        x->finalCheck = false;
+        InvokeEntry* entry2 = lookupInvokeEntry(x, argsKey, argsTempness, failures);
+        if (entry2 != NULL)
+            matchFinalError(failures);
+        x->finalCheck = true;
+    }
```
jckarter, line 1841: Couldn't this be done entirely in lookupInvokeEntry? Search through the candidates, and if you find a matching non-final overload, return that overload; if you find a matching final overload, save it off and continue searching; if you find another final overload, return an error; if you reach the end, return the final overload, if any.

PR #421, opened 2012-10-15
algorithms.clay: add join with no separator (concat)

galchinsky:
join("", seq) is good, but somethimes there's no way to create an empty sequence


jckarter:
`join` with no separator is `cat`.

jckarter:
Oh, I see, you want to concat a sequence of sequences. Looks good.

Issue #90, opened 2011-12-25
Lambdas that capture environment from alias function generate invalid LLVM


jckarter:

jckarter:
Partially fixed by @agemogolk in #425. Local aliases (viz. `test/lib-clay/lambdas/alias/2`) and multiple-value alias bindings still need work.

jckarter:
Fixed by @agemogolk in #429

PR #425, opened 2012-10-17
Enable lambda capture of alias function arguments - fix issue #90

ghost:
Enabled test for issue #90


jckarter:
Excellent! #90 was one I was never really sure how to approach, but this should work. Does your fix work for an unpacked multiple-value alias, such as `..a`? Can test `lambdas/alias/2` be enabled now?

ghost:
Unfortunately this only handles the case of arguments. Alias bindings in the captured env still produce an alias analysis bug e.g. lambdas/aliases/3 gives:

```
clay: BasicAliasAnalysis.cpp:480: virtual llvm::AliasAnalysis::AliasResult {anonymous}::BasicAliasAnalysis::alias(const llvm::AliasAnalysis::Location&, const llvm::AliasAnalysis::Location&): Assertion `notDifferentParent(LocA.Ptr, LocB.Ptr) && "BasicAliasAnalysis doesn't support interprocedural queries."' failed.
Stack dump:
0.  Running pass 'CallGraph Pass Manager' on module 'main-disabled.clay'.
1.  Running pass 'Global Value Numbering' on function '@"(() => { println(x); })() clay"'
```

I don't have time now, but i'll check and add a test (if it works!) for the multi-value case this evening.

ghost:
Actually quick look shows this won't work, think it just needs a small tweak to do analyzeMulti etc.

jckarter:
OK, well what you have looks good for now.

PR #429, opened 2012-10-20
Enable lambda capture of local alias & var arg expressions - issue #90

ghost:
Modified checkForeignExpr to handle alias bindings & var args.
Moved lambda operator code to core.operators.
Enabled lambda/alias tests.
Added lambda/alias/vararg test.


jckarter:
Looks good. Nice work!

PR #430, opened 2012-10-21
maybe, list and state monads with examples

galchinsky:
I wanted to test a typesystem if there was a way to write generic functions that work with types like monads. I think that the result is good and can be pulled, but maybe there's another way to solve it.
Problems:
1. Type propagation is only unidirectional, so you can't write a generic function  `[S, A] mreturn (a : A) : S[A] = S(A)` like in Hindley–Milner languages: return type is unbound. A workaround is to get help from the caller:

```
[S, A] overload mreturn (a : A, #S) = S(A);
overload mreturn(a : A, #Maybe[T])= Maybe[A](a);
```

You may look at the tests in this commit to be sure that it doesn't look very ugly.
2. sequence function should be `sequence(ms : Vector[M[A]]) : M[Vector[A]]` in HM-language.
In clay it is `[M when Monad?] sequence(ms : Vector[M])`. `M` is a black box. We need to extract `A` type from `M` and to change M[A] to M[Vector[A]]. This extraction can be done with 2 helper functions for every box (Maybe, Vector, State or any other).

```
[T]overload typeInBox(#Maybe[T]) = T;
[T, U] overload changeTypeInBox(#Maybe[T], #U) = Maybe[U];
```

It can be automated by the compiler, but I'm not sure that it is good idea: for example, State[S, A] has 2 parameters and you need to change only the second.
then we can write `sequence` this way:

```
[M when Monad?(M)]
sequence(ms: Vector[M]){
    alias A = typeInBox(M);
    alias R = changeTypeInBox(#M, Vector[A]);//result type
    //...
}
```
1. Bind function should be `mbind(f : Function[A, MB], x : MA) : MB`. But lambdas needs to be casted to a Function object, so I think it's better to write `[MA when Monad?(MA)]mbind(f, x:MA)`. Resulting type is unknown again. It can be extracted via `alias MB=Type(f())` and we already have `typeInBox` to extract B. I needed this extraction only in `State` `mbind`.
2. `With` syntax seems to be broken, especially when it is nested. And in fact it is more verbose than straight operator `>>=` using. I think it's better to make a classic do-notation.
3. State monad needs capture-by-value `=>` lambdas to work properly, so it is everywhere in generic functions.


jckarter:
The `with` notation is backwards because of the lack of support for single bindings after variadic bindings at the time it was implemented. Now that `foo(..a, b)` works, it would make more sense for `with` to desugar that way, and for `mbind` to take its arguments in the traditional order.

Handling lambdas in `mbind` is probably best by casting the lambda to Function in a default implementation, rather than parameterizing the monad type on the lambda. Aside from opaque types, parameterizing on the lambda also creates extra generic specializations.

```
define mbind(m, f);
[M when Monad?(M)] overload mbind(m:M, f) = mbind(m, Function[[typeInBox(M)], [changeTypeInBox(M, CallOutputTypes(f))]](f));

overload mbind(m:SomeMonad[A], f:Function[[A], [MB]]) = /* specific implementation goes here */;
```

galchinsky:
Casting in a default implementation makes everything more clear. Anyway I think lambda->function object interaction could be better.
I also added Dist monad example.

test/example/distmonad/main.clay
```diff
@@ -0,0 +1,117 @@
...
+
+var driver = freqs(reverse(Vector(array([0.2, Cautious], [0.6, Normal], [0.2, Agressive]))));
+
+simulate(d1, d2, light) = actOn(d1, light) >>=( a1 =>
+                          actOn(d2, light) >>=( a2 =>
+                          collision(a1, a2)));
+
+simulateOverDrivers(light) = driver >>=( d1 =>
+                             driver >>=( d2 =>
+                             simulate(d1, d2, light)));
```
jckarter, line 107: The precedence of lambda arrows should be such that you can elide the parens, as in `driver >>= d1 => driver >>= d2 => simulate(...`

lib-clay/core/operators/operators.clay
```diff
@@ -715,10 +715,33 @@ forceinline MonoType?(F) : Bool = Mono?(typeToRValue(F));
 [F when MonoType?(F)]
 alias MonoOutputTypes(f:F) = ..CallOutputTypes(f, ..MonoInputTypes(f));
 
+/// @section functors and monads
+
+
+public define typeInBox(T);
+public define changeTypeInBox;
```
jckarter, line 722: In `sequences.force` there are functions `UnderlyingContainerConstructor` and `ForcedContainerConstructor`, whose purpose is very similar to these—`UnderlyingContainerConstructor` gives the parameterized type symbol of a container type without the parameters, and `ForcedContainerConstructor` gives a new container type that's intended to be like another container type but with a different element type. It'd be cool if these protocols could be combined into one.

test/example/distmonad/main.clay
galchinsky, line 107: It gave me a parse error at (=>) at very big and very small precedences both. It's about "=>" precendence I think if I didn't do something wrong

jckarter, line 107: That sounds like a bug. I'll take a look.

jckarter, line 107: You're right—currently it only parses lambdas at the top level of an expression. It should be possible to lift this restriction, though it will make parsing slower because more subexpressions will need to be tried as lambdas.

galchinsky, line 107: I think that it's not very important if we implement do-notation, so have left this braces in the code

lib-clay/core/operators/operators.clay
galchinsky, line 722: I think it would be nice to have type parameters introspection (somehow you proposed `[T, U] overload changeTypeInBox(#T, #U) = SymbolRoot(T)[U];`):
I think it should be UnpackType(V[Function[[A],[B]]) primitive which returns something like [V, [Function, A, B]].
Could you describe shortly where and how primitives defined in compiler? Right now I don't understand where they are and their interaction with Clay code.

Then the protocol would be `boxType` and  `typeInBox` with trivial default implementations and overloads if user needs support for types like MyArray[n: Int, A]. After this `sequences` may be patched to use more general `boxType` and  `typeInBox`.

jckarter, line 722: Primitive implementation is a bit diffuse unfortunately—searching the compiler dir for `PRIM_` will give you a sense of where everything is. `loader.hpp` defines an `enum PrimOp` with identifiers for all of the primitives, and `loader.cpp` contains a function `makePrimitivesModule` which populates the `__primitives__` module with these symbols. The actual behavior of the primitives is implemented in `analyzer.cpp`, `codegen.cpp`, and `evaluator.cpp`. `analyzePrimOp` performs type propagation for primitives. `codegenPrimOp` generates LLVM code for runtime implementation of primitives. `evalPrimOp` evaluates primitives at compile time.

galchinsky:
Looked more carefully. Puporse of `UnderlyingContainerConstructor` differs from mine. `UnderlyingContainerConstructor` returns a constructor, and it's Vector by default (so you can send some lazy seq to Vector constructor in `force` function). It needs to be declared only for strict sequences, for lazy sequences it's recursive up to strict. My functions don't return a constructor, they return a type, and they need to be declared for every type.
`typeInBox(#Sliced[Sliced[Vector[Int]]]) == Sliced[Vector[Int]]`, but `UnderlyingContainerConstructor(#Sliced[Sliced[Vector[Int]]]) =  ...recursion... = Vector`
I've renamed UnderlyingContainerConstructor and added this three functions to `typeparam` module (https://github.com/galchinsky/clay/commit/2482d88a317c7feb119efc07f045f4edf403cd1b), but it doesn't look well and I think it's not good to mix UnderlyingContainerConstructor with typeInBox.

galchinsky:
I mean it doesn't have any contact except some type magic

jckarter:
Well, the reason I brought it up is that ultimately what `force` is trying to accomplish is to derive a `C[U]` type from a `C[T]` input, which could be accomplished with your protocol as well. One thing I was thinking was that if root symbol pattern matching were supported, then most generic containers could be matched by a `M[A, ..OtherParams]` pattern. The only types this would miss out would be monomorphic containers or other weird containers that didn't express the internal type as a type parameter.

Issue #135, opened 2012-01-04
Explicit lamba capture clauses, stateless lambdas


jckarter:
It could be useful to allow environment captures to be made explicit in lambda syntax, if we can come up with a syntax that isn't too C++0x-ish. Having an arrow form to assert a stateless lambda would also be useful in contexts where a stateless function must be used.

ghost:
Possible extension to lambda syntax:

```
var y, z = 6, 7;
x => x + ref y + z  // y captured by ref, z is copied free variable
x -> x + ref y + z  // ref y is a no-op, all captured by ref
```

`->` guarantees stateless. I think this would cover all cases. Any thoughts?

Note that I have assumed here that capturing by ref and value by the same lambda is in some way useful, it occurs that this may not be the case and the current lambda syntax is sufficient. In which case I guess this issue is no longer an issue.

jckarter:
I think that would clash with the `return ref` syntax. I don't see how `->` guarantees stateless—your example still captures `y` and `z`. I'm not sure how generally useful mixed capture is, but stateless lambdas have the ability over other lambdas to be turned into stateless `CodePointer`s and `CCodePointer`s, which is sometimes important when interfacing with C functions, and explicit capture is useful for verification or documentation purposes, if there's a local that really shouldn't be captured or something like that.

ghost:
I guess I was falsely under the impression that 'stateless' included capturing by reference, just no copies. My mistake.

jckarter:
Technically you're right (the best kind of right), but by "stateless" I really meant "representable as a single function pointer".

ghost:
I suppose the arrow syntax could be extended with `~>` for stateless. Not a very clear distinction from the current syntax though.

jckarter:
Yeah, that's maybe the least ugly choice. Unfortunately I think explicit capture clauses are just inherently ugly; if we free up the colon by removing the trailing block sugar, one possibility would be something like `args ~>:(a, ref b) body`.

Another interesting case to support would be capture-by-move, which is necessary to support capture of move-only objects like UniquePointer, and is also a good optimization for things like the monad pattern, where the lambda is the last consumer of the scope and so can move it into itself.

ghost:
[...]

PR #431, opened 2012-10-22
Add stateless lambda ~>, partial fix for issue #135

ghost:
Add stateless lambda form "~>", raises error when attempting to add free variables.
Add test for 'capture within stateless' failure.


jckarter:
Looks good, thanks.

PR #432, opened 2012-10-23
Repl: catch CompilerError instead of std::expection

galchinsky:
Repl: catch CompilerError instead of std::expection
Hide -fexceptions warnings in VS

Still bugs:
- SIGABRT recovery in MSVS is sill buggy. MSDN says that you can recover with longjmp, but after one catched SIGABRT another starts the debugger. `__except` can't catch SIGABRT
- Noticed not full recovery after exception, perhaps some module state becomes undefined after an error:

```
g:/clay/build/compiler>clay ..\..\examples\factorial.clay -repl
clay>factorial1(123123123123);
(1,11): error: int32 literal out of range
clay>factorial1(1);
(1,11): error: int32 literal out of range
```

factorial1(1) is not out of range


jckarter:
Looks good, thanks. IIRC the equivalent MSVC flag is `/EHsc`.

PR #434, opened 2012-10-26
fix build fail because of int-enum conversion

galchinsky:
My gcc 4.4.7 (Debian 4.4.7-2) fails with default cmake parameters

### DIFF
--- a/compiler/env.cpp
+++ b/compiler/env.cpp
@@ -117,10 +117,11 @@ static void getIntrinsicSymbols(ModulePtr module)
 #include "llvm/Intrinsics.gen"
 #undef GET_INTRINSIC_NAME_TABLE
     };
-    for (llvm::Intrinsic::ID id = llvm::Intrinsic::ID(1);
-         id < llvm::Intrinsic::num_intrinsics;
-         ++id)
+    for (size_t idIter = llvm::Intrinsic::ID(1);
+         idIter < llvm::Intrinsic::num_intrinsics;
+         ++idIter)
     {
+        llvm::Intrinsic::ID id = static_cast<llvm::Intrinsic::ID>(idIter);
         llvm::SmallString<64> identifier;
         getIntrinsicIdentifier(&identifier, intrinsicNames[id]);
         module->publicSymbols[identifier].insert(


compiler/env.cpp
```diff
@@ -117,10 +117,11 @@ static void getIntrinsicSymbols(ModulePtr module)
 #include "llvm/Intrinsics.gen"
 #undef GET_INTRINSIC_NAME_TABLE
     };
-    for (llvm::Intrinsic::ID id = llvm::Intrinsic::ID(1);
```
jckarter, line 120: You can rewrite this as a static_cast rather than casting it on a separate line in the loop.

galchinsky, line 120: If you mean 'id = static_cast<size_t>(llvm::Intrinsic::ID(1));', then I'll have to cast it in 'new IntrinsicSymbol', so it won't be shorter

jckarter, line 120: That shouldn't be necessary. I think gcc is trying to apply the C++11 const initialization promotion rules (even though it's not C++11), which don't apply to runtime expressions.

galchinsky, line 120: With static_cast in `for` it is

```
cd ... && /usr/bin/c++    -g   -I/usr/local/include  -D_DEBUG -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -g -fvisibility-inlines-hidden -fno-exceptions -fno-rtti -fPIC -Woverloaded-virtual -Wcast-qual -O0 -UNDEBUG -fexceptions -o CMakeFiles/compiler.dir/env.cpp.o -c clay/compiler/env.cpp
clay/compiler/env.cpp: In function ‘void clay::getIntrinsicSymbols(clay::ModulePtr)’:
clay/compiler/env.cpp:127: error: invalid conversion from ‘size_t’ to ‘llvm::Intrinsic::ID’
clay/compiler/env.cpp:127: error:   initializing argument 3 of ‘clay::IntrinsicSymbol::IntrinsicSymbol(clay::Module*, clay::IdentifierPtr, llvm::Intrinsic::ID)’
```

It fails here even without any flags except -c and -o

jckarter, line 120: What does the code look like? The type of `id` should be `llvm::Intrinsic::ID`, not `size_t`.

galchinsky, line 120: 1. Your code => `env.cpp:122: error: no match for ‘operator++’ in ‘++id’` (without any flags or with default cmake falgs)
2. `for (size_t id = static_cast<size_t>(llvm::Intrinsic::ID(1));` => error in my previous message.

If i change the line with error to `new IntrinsicSymbol(module.ptr(), Identifier::get(identifier), static_cast<llvm::Intrinsic::ID>(id)));` in compiles, but I don't think that it is better than the approach in the current commit

jckarter, line 120: This doesn't work?

```
for (llvm::Intrinsic::ID id = static_cast<llvm::Intrinsic::ID>(1);
    /* everything else same as before */
```

If not, that sounds like a GCC bug.

galchinsky, line 120: The error is the same as in original code : `122: error: no match for ‘operator++’ in ‘++id’`. It doesn't like increment on enums.

jckarter, line 120: I see. In that case, your fix seems best.

PR #443, opened 2012-11-11
Set clang include dir in clay-bindgen build - fix issue #419

ghost:
Clang lib tooling states the default dir should be used.
Have passed the dir into the clay-bindgen build via CMake.
This works on Archlinux, but could do with testing on other platforms.

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -12,17 +12,20 @@ else (WIN32)
     set(LIBCLANG_LDFLAGS -L${LLVM_LIBDIR} -Wl,-rpath,${LLVM_LIBDIR} -lclang)
 endif (WIN32)
 
+
 if(BUILD_BINDGEN)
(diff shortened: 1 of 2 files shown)


galchinsky:
works fine now

jckarter:
I guess that works. It'd be nice to also figure out why libclang doesn't do this by default on Linux and fix it upstream too.

ghost:
This issue has already been fixed upstream (clang 3.2 trunk).
Probably not worth merging this temporary fix if we are close to upgrading llvm/clang.

jckarter:
Since LLVM 3.2 just forked a few days ago and will be released soon, it'd be a good time to port over to the new version. There are a couple other LLVM bug workarounds that could then be backed out too.

jckarter:
Closing this until Clay is updated to use LLVM 3.2. If it's still a problem after that we can reopen.

PR #444, opened 2012-11-16
Extract  parser.hpp from clay.hpp. Multi-line input for repl

galchinsky:
`next` token function in parser runs a callback to get tokens from console if it is in repl

```
$clay -repl ../examples/factorial.clay
clay>[T]   
foo(z:T){
var x = 5;
var y = 6;
println(x * y * z);
}
clay>foo(1);
30
```

`Parser.hpp` extraction is for using `#include lexer.hpp` in the header.

### DIFF
--- a/compiler/parser.cpp
+++ b/compiler/parser.cpp
@@ -9,7 +9,6 @@ map<llvm::StringRef, IdentifierPtr> Identifier::freeIdentifiers;
 static vector<Token> *tokens;
 static int position;
 static int maxPosition;
-static int sourceSize;
 static bool parserOptionKeepDocumentation = false;
 
 static AddTokensCallback addTokens = NULL;
@@ -3114,7 +3113,7 @@ static bool module(llvm::StringRef moduleName, ModulePtr &x) {
 static bool replItems(ReplItem& x, bool = false) {
     inRepl = false;
     int p = save();
-    if (expression(x.expr) && position == sourceSize) {
+    if (expression(x.expr) && position == tokens->size()) {
         x.isExprSet = true;
         return true;
     }
@@ -3173,7 +3172,6 @@ void applyParser(SourcePtr source, int offset, int length, Parser parser, Parser
 
     tokens = &t;
     position = maxPosition = 0;
-    sourceSize = t.size();
 
     if (!parser(node, parserParam) || (position < (int)t.size())) {
         Location location;
--- a/compiler/parser.cpp
+++ b/compiler/parser.cpp
@@ -3110,13 +3110,7 @@ static bool module(llvm::StringRef moduleName, ModulePtr &x) {
 // REPL
 //
 
-struct REPLitem {
-    vector<TopLevelItemPtr> toplevels;
-    vector<ImportPtr> imports;
-    vector<StatementPtr> stmts;
-};
-
-static bool replItems(REPLitem& x, bool = false) {
+static bool replItems(ReplItem& x, bool = false) {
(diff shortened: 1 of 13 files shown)


compiler/parser.cpp
```diff
@@ -11,9 +11,28 @@
 static int maxPosition;
 static bool parserOptionKeepDocumentation = false;
 
+addTokensCallback addTokens = NULL;
```
jckarter: Is there any way to avoid the global state here, maybe by factoring out the parser state into a Parser object?

compiler/parser.hpp
```diff
@@ -0,0 +1,32 @@
...
+void parseStatements(SourcePtr source, int offset, int length,
+    vector<StatementPtr> &statements);
+void parseTopLevelItems(SourcePtr source, int offset, int length,
+    vector<TopLevelItemPtr> &topLevels, Module *);
+void parseInteractive(SourcePtr source, int offset, int length,
+                      vector<TopLevelItemPtr>& toplevels,
+                      vector<ImportPtr>& imports,
+                      vector<StatementPtr>& stmts);
+
+typedef vector<Token>(*addTokensCallback)();
```
jckarter: The type name should be capitalized.

jckarter:
Sorry about not looking at this sooner. The changes look good to me aside from the comments I made in the patch.

compiler/parser.cpp
galchinsky: I see I've forgotten `static` keyword here. Do you mean moving from c-style singleton modules to classes? I think if to do it, it should be done for every module. On the other side, I didn't have any troubles with it except this https://github.com/jckarter/clay/blob/master/compiler/codegen.cpp#L7209 global-ctors hack after your explanation

galchinsky:
Added static keyword (comment#1), capitalized typename and rebased against master

galchinsky:
If input line can be parsed as an expression, print it:

```
clay>five() = 5;
clay>5 + five() //note there's no semicolon
10
```

galchinsky:
add 2 repl commands: modules, globals

```
clay>var x = 5; var y = 7;
clay>:globals
x : Int32 : 5
y : Int32 : 7
clay>x = x + y;

clay>:globals
x : Int32 : 12
y : Int32 : 7
```

chetankothari:
Hey Dmitry Galchinsky, can you give me some guidance on how to contribute to REPL part of clay. I'am new to Github as well as to clay as well. But i am wiling to contribute to repl part of clay.

galchinsky:
1. Have you worked with git? If you didn't, start with any of git tutorials. 
2. Clone this repo (`git clone git://github.com/jckarter/clay.git`)
3. Read README and try to build the compiler. It may take some time to build LLVM and clang. Note that today you need LLVM-3.1, not 3.2.
4. If everything is ok, click `fork` to get your own copy of repo and clone it. 
5. Enjoy. Repl is started by `clay -repl` or `clay -repl somefile.clay` in console. The entry point is`runInteractive()` in `compiler/interactive.cpp`. I'll try to write issues list when you get here.

chetankothari:
Thanks a ton Dmitry Galchinsky, wil try to do as you said and will get back
to you as soon as i am comfortable using Github and understand sufficient
of clay. Thanks once again.

On Mon, Nov 26, 2012 at 1:42 PM, Dmitry Galchinsky <notifications@github.com

jckarter:
Looks good @galchinsky. Looking forward to contributions from you, @chetankothari !

PR #445, opened 2012-11-19
Remove complex types from compiler/core - use newtype instead

ghost:


jckarter:
This would make sense, except that it breaks abi compatibility with C `_Complex` types. `Imag` could at least be a newtype of `Float` though.

ghost:
It's preferable to have `Imag` as default in the compiler due to use of imaginary literals. Maybe it would be possible to implement literal suffixes as postfix operators . . . not too sure how that would work though.

This was really an experiment to see if it would work and provide a test for the newtype code. If there's no way around the abi issue i'll just apply the bug fixes.

ghost:
Pushed newtype fixes. Close pull-request due to abi issue mentioned above.

PR #446, opened 2012-11-19
Support variadic fields in Records

ghost:


compiler/clay.hpp
```diff
@@ -2828,6 +2835,9 @@ struct RecordType : public Type {
 
     vector<IdentifierPtr> fieldNames;
     vector<TypePtr> fieldTypes;
+    unsigned varFieldPosition;
+    bool hasVarField:1;
```
jckarter: To minimize size, I sorted all the class fields by descending alignment order, so to keep that going, these two fields should probably go to the end of the object.

jckarter:
Looks good. You could potentially also support multiple variadic fields, although you wouldn't be able to use the default constructor with them.

Issue #448, opened 2012-11-23
lib-clay's opengl incorrectly declares "initGLFunctions"


alfredr:
lib-clay's opengl fails to declare the overloadable function "initGLFunctions" with define. As a result the opengl examples will no longer compile.

jckarter:
Fixed in #449.

PR #449, opened 2012-11-23
declare initGLFunctions with "define" (this fixes issue #448)

alfredr:
This pull request closes issue #448.


jckarter:
Nice catch. Thanks!

PR #454, opened 2012-12-02
Compiler: fix compilation with -g

Blei:
The LLVM `expect` intrinsic only works with integer types, but we used
it with pointer types when checking for exceptions. Add some bitcasts
(which should be optimized to nothing) to stop LLVM from throwing
assertion failures.

### DIFF
--- a/compiler/codegen.cpp
+++ b/compiler/codegen.cpp
@@ -2620,21 +2620,25 @@ void codegenLowlevelCall(llvm::Value *llCallable,
     if (!ctx->checkExceptions)
         return;
     llvm::Value *noException = noExceptionReturnValue();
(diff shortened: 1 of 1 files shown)


ghost:
The following code fails to compile using `-g` with this patch (segfaults):

```
var ..a = 3, "test";
println(..a);
```

looks like this bug may have been masking a bug in my binding debug info code.

Blei:
Pushed a patch that fixes that segfault. Not sure if there are still remaining segfaults wating to happen, but the tests still compile.

ghost:
Thanks, Blei. 

Note: Might be worth running the test suite with the `-g` flag . . .

Issue #102, opened 2011-12-31
Shorthand syntax for conversions in overloads


jckarter:
A common boilerplate pattern is a pair of overloads, one that converts values to a canonical type and another that actually applies an algorithm to the canonical type:

```
foo(x, y) = foo(Int(x), String(y));

overload foo(x:Int, y:String) { ... }
```

It would be nice to have this built into the language. An idea KS and I came up with was to reuse the `as` keyword. The following definition:

```
foo(x as Int, y as String) { ... }
```

would be equivalent to the following two definitions:

```
foo(x, y) = forward ..foo(x as Int, y as String);

overload foo(x:Int, y:String) { ... }
```

The new `as` operator should be distinct from the constructor function, since operations like `int as Int` should be no-ops, and the implicit conversion operation should be limited to cheap and safe value-to-argument conversions in keeping with the "explicit is better than implicit" philosophy.

ghost:
How about:

``` text
  foo(x:TX as Int, y:TY as String) {
      /* body of foo */
  }
```

desugers as:

``` text

  [TX, TY | Convertible?(TX, Int) and Convertible?(TY, String)]
  inline foo(forward x: TX, forward y: TY) = ...forward foo(TX, TY, x as Int, y as String);

  foo(TX, TY, x: Int,  y: String) {
      /* body of foo */
  }
```

Thus allowing access to the original types in the body of foo. This would also look nicer when procedures allow static parameters. Would this work?

jckarter:
@agemogolk That makes sense. A definition with `as` could then apply a predicate or pattern restriction to the input types prior to conversion too, such as `[T | Integer?(T)] foo(x:T as Int) { ... }`.

ghost:
Implementing this as an overload de-sugar could be problematic with respect to overload resolution and final overloads. Maybe it would be better to integrate argument promotion into the, yet to be decided, more sophisticated resolution scheme. Provide an overload-able `Convertible?` primitive etc. Although this would probably end up being similar to VB overloads . . . not sure how well that would work. 
Also, I suppose i'm not entirely convinced that new syntax should be introduced for a pattern that could be handled with  predicates (or type-classes) e.g.

```
[T, S  when Convertible?(T, Int) and Convertible?(S, String)]
foo(x:T, y:S) {
    var i, s = Int(x), String(y);
    /* body of foo */
}
```

jckarter:
[...]

PR #461, opened 2012-12-20
'as' expression and argument conversions (fix issue #102)

ghost:
Implement 'as' expression:
`x as Int64`
desugars too:
`alias overload asExpression(value, #T) = T(value);`
The type consructor overload is a quick default and should probably be changed.

'as' argument conversions:

```
[T when Integer?(T)]
overload foo(x:T as Int64, y as Float64) = T, x;
```

desugars too:

```
[T when Integer?(T)]
overload foo(forward x:T, forward y) = 
    forward ..foo(T, x as Int64, y as Float64);

[T]
overload foo(#T, x:Int64, y:Float64) = T, x;
```

Add couple of simple tests.


jckarter:
This looks OK at a quick glance. Is there a reason you closed it?

stepancheg:
I'm opposing the 'as expression' part (it provides little benefit, but complicates the language). 'as argument conversion' in parameter is nice.

ghost:
I've closed this issue temporarily for a couple of reasons. Currently the 'as' desugars are done during loading, it works ok, but I'm going to try to desugar on demand. Also the commit history is a mess, i'll tidy this up a bit before reopening.

ghost:
Tidied up the code/commits. I had a quick go at integrating the desugar into the match system without success. Seems to work fine as it is.

jckarter:
Sounds good. In my thinking, the 'as' expression isn't as important as the sugar for conversions, since the alternatives are either excessive instantiation or the boilerplate of having to write two overloads.

ghost:
Removed `as` as an expression/operator as it's not completely consistent with argument use.

PR #462, opened 2012-12-31
update for llvm 3.2: master branch

crabtw:
This pull request is for master branch

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,8 +32,8 @@ if(UNIX)
         message(FATAL_ERROR "${LLVM_CONFIG} failed")
     endif()
 
-    if(NOT ${LLVM_VERSION} MATCHES "^3.1(svn)?$")
-        message(FATAL_ERROR "Clay requires LLVM 3.1.")
+    if(NOT ${LLVM_VERSION} MATCHES "^3.2(svn)?$")
+        message(FATAL_ERROR "Clay requires LLVM 3.2.")
     endif()
 
     execute_process(
--- a/README.txt
+++ b/README.txt
@@ -14,7 +14,7 @@ is `v0.1`; for normal use you will want to switch to that branch:
 Pre-requisites
 --------------
 
-Clay requires CMake version 2.6 or later, and LLVM and Clang 3.1. Note that
+Clay requires CMake version 2.6 or later, and LLVM and Clang 3.2. Note that
 LLVM versions are not backward-compatible so developer versions of LLVM and/or
 later releases are not compatible with this version of Clay.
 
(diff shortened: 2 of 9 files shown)


jckarter:
Looks good. I had updated Clay for LLVM 3.2 locally but ran into some externals/ test failures I didn't have time to investigate. Do all the tests pass for you?

crabtw:
Two tests do not pass. I have no idea how to fix them.
- lib-clay/externals/abi/common
  Some arguments become `undef`, for example, `call void @c_2(%"x86-64 Struct2" undef)`.
- lib-clay/llvm/intrinsics
  `Intrinsics.gen` does not have verifier anymore, so the `ia.outOverloadedTypes` is always empty.
  That causes `llvm::Intrinsic::getDeclaration` assertion failed.

jckarter:
It looks like LLVM 3.2 is stricter about `bitcast`ing from a shorter to a longer type. Bitcasting `{i32}` to `{i64}` now evaluates to a poison value, which was breaking Clay's x86-64 ABI implementation. Here's a patch that makes Clay generate a form-fitting word for the ABI struct that fixes the externals tests for me on OS X:

``` diff
diff --git a/compiler/externals.cpp b/compiler/externals.cpp
index bb35e36..d4a666a 100644
--- a/compiler/externals.cpp
+++ b/compiler/externals.cpp
@@ -878,21 +878,26 @@ llvm::Type *X86_64_ExternalTarget::llvmWordType(TypePtr type)
     llvm::StructType *llType = llvm::StructType::create(llvm::getGlobalContext(), "x86-64 " + typeName(type));
     vector<llvm::Type*> llWordTypes;
     WordClass const *i = wordClasses.begin();
+    size_t size = typeSize(type);
     while (i != wordClasses.end()) {
+        assert(size > 0);
         switch (*i) {
         // docs don't cover this case. is it possible?
         // e.g. struct { __m128 a; __m256 b; };
         case NO_CLASS:
             assert(false);
             break;
-        case INTEGER:
-            llWordTypes.push_back(llvmIntType(64));
+        case INTEGER: {
+            size_t wordSize = size >= 8 ? 64 : size*8;
+            llWordTypes.push_back(llvmIntType(wordSize));
             ++i;
             break;
+        }
         case SSE_INT_VECTOR: {
             int vectorRun = 0;
             do { ++vectorRun; ++i; } while (i != wordClasses.end() && *i == SSEUP);
-            // 8-byte int vectors are allocated to MMX registers
+            // 8-byte int vectors are allocated to MMX registers, so always generate
+            // a <float x n> vector for 64-bit SSE words.
             if (vectorRun == 1)
                 llWordTypes.push_back(llvm::VectorType::get(llvmFloatType(64), vectorRun));
             else
@@ -946,6 +951,8 @@ llvm::Type *X86_64_ExternalTarget::llvmWordType(TypePtr type)
             assert(false);
             break;
         }
+        assert(size >= 8 || i == wordClasses.end());
+        size -= 8;
     }
     llType->setBody(llWordTypes);
     return llType;
```

Fixing the intrinsics module will probably be more annoying. Do you know if the input and output type information for LLVM intrinsics got moved somewhere else that can be used by Clay?

crabtw:
You can use `getIntrinsicInfoTableEntries` to get the information (https://gist.github.com/4432544#file-intrinsics-diff-L121) and the first entry of table is for output type.

It uses `Verifier::VerifyIntrinsicType` to verify intrinsic types (https://gist.github.com/4432544#file-intrinsics-diff-L30).

jckarter:
I'll merge this and put up another ticket to update the intrinsics module for LLVM 3.2. Thanks @crabtw !

PR #470, opened 2013-01-09
pass optimization level to TargetMachine correctly

crabtw:
Now there is no optimization level for `TargetMachine::addPassesToEmitFile`.
This patch moves the level to `Target::createTargetMachine`.


jckarter:
Good catch. Patch looks good to me.

ghost:
Thanks @crabtw.

Issue #190, opened 2012-01-19
"loss of data" warnings casting size_t to int


jckarter:
Building for win64 brings up a lot of these, which should all be fixed.

PR #471, opened 2013-01-09
Fix implicit conversion warnings

ghost:
Fixed most warnings from -Wconversion, should fix issue  #190


jckarter:
Looks good, though I don't have a win64 machine handy to verify that #190 was fixed.

PR #478, opened 2013-01-13
Final overloads

ghost:
Implemented final overloads as discussed in #464, #383, #334.
- Overloads defined using the 'overload' keyword are final.

```
    define foo;
    overload foo(x) {}     // foo(x:T) is final
    overload foo(y:Int) {} // ambiguous match error
```
- Overloads defined using the 'default' keyword can be overridden by subsequent 'default' definitions and 'overload's.

```
    define foo;
    default foo(x) = false;
    default foo(x:Int32) = true;    //OK as foo(x:T) not final
    overload foo(x:Float32) = true; //OK as foo(x:T) not final
    default foo(x:Int32) = false;  //OK as foo(x:Int32) not final
```
- Default overloads may be defined in any module.
- Standalone functions are unchanged.

Final overloads are switched off by default to allow the library to be incrementally updated. A compiler flag '-final-overloads' has been added to turn final overloads on.

The library has been updated with 'default' overloads to allow the bindgen tool and hello world example to compile with final overloads enabled.

Some simple tests have been added using the -final-overloads flag.

Issues:

The ambiguous match error output is currently untidy and possibly too verbose.


compiler/error.cpp
```diff
@@ -305,6 +305,30 @@ void argumentError(size_t index, llvm::StringRef msg) {
...
+        llvm::errs() << match2->overload->location.source->fileName
+            << '(' << line+1 << ',' << column << "): " << "error: ambiguous overload\n";
+        llvm::errs().flush();
+        
+    }
+    string buf;
+    llvm::raw_string_ostream err(buf);
+    err << "ambiguous overloads for call-site\n";
+    
+    error(err.str());
```
jckarter: It'd make more sense to write everything to the error string. IIRC that's how the 'no matching overload' error message works.

jckarter:
Looks good to me.

ghost:
Ambiguous match error reporting integrated into match failure system.
This code:

```
define foo;
overload foo(x) = true;
overload foo(y:Int32) = false;

main() {
    foo("bar");
    foo(2);
}
```

produces this error (with -final-overloads):

```
main() {
    foo("bar");
    foo(2);
-----------^
}
../test/lang/overloads/final/ambiguous/main.clay(7,4): error: call matches ambiguous overloads
    ../test/lang/overloads/final/ambiguous/main.clay(3,0)
        matched
    ../test/lang/overloads/final/ambiguous/main.clay(2,0)
        matched

compilation context: 
    foo(Int32)
  ../test/lang/overloads/final/ambiguous/main.clay(7,4):
    main()
  /home/jeremy/dev/clay/clay/build/compiler/../../lib-clay/core/system/system.clay(23,30):
    getExitCode()
  /home/jeremy/dev/clay/clay/build/compiler/../../lib-clay/core/system/system.clay(23,26):
  /home/jeremy/dev/clay/clay/build/compiler/../../lib-clay/core/system/system.clay(39,13):
    external main
```

test/lang/overloads/final/ambiguous/main.clay
```diff
@@ -0,0 +1,8 @@
+define foo;
+overload foo(x) = true;
+overload foo(y:Int32) = false;
+
+main() {
+	foo("bar");
+	foo(2);
+}
```
stepancheg: Test does not check which one of these two calls is ambiguous overload. I assume first call is good, and second is failure. But test would incorrectly succeeds even if first call fails. I think test should be split into two tests.

test/lang/overloads/final/default/test.clay
```diff
@@ -0,0 +1,12 @@
...
+default foo(x, y) = true;
+overload foo(x:Int32) = false;
+overload foo(x:Char, y:Byte) = false;
+
+main() {
+	foo("bar");
+	foo(2, "bar");
+	foo(2);
+	foo('a', 2ss);
+}
```
stepancheg: This test could check which overload is actually used. Something like this:

```
import printer.*;

define foo;
default foo(x) = 1;
default foo(x, y) = 2;
overload foo(x: Int32) = 3;
overload foo(x: Char, y: Byte) = 4;

main() {
    println(foo("bar"));
    println(foo(2, "bar"));
    println(foo(2));
    println(foo('a', 2ss));
}
```

so test.clay should be renamed to main.clay, and out.txt need to be created.

ghost: Sure. I'll update it.

stepancheg:
@agemogolk I like this pull request: language change is good, and implementation is also good.

ghost:
Thanks @stepancheg . I'll just wait and see if @jckarter has any further comment on error messages before merging.

jckarter:
Reusing the match failure error code is a good idea. Looks good to me.

ghost:
I find the wording quite ambitious. The word default screams for another overload, and the word overload doesn't imply a final. What is so special about the word overload that you stick to it? What about apply/final(ize), apply/set, apply/fix or apply/define? overload/final is more clear too.

PR #479, opened 2013-01-13
Fix __intrinsics__ module for LLVM 3.2

### LINKED ISSUE #466: Update __intrinsics__ module implementation for LLVM 3.2
LLVM 3.2 changed how LLVM intrinsics are registered. Per @crabtw:

> You can use getIntrinsicInfoTableEntries to get the information (https://gist.github.com/4432544#file-intrinsics-diff-L121) and the first entry of table is for output type.
> 
> It uses Verifier::VerifyIntrinsicType to verify intrinsic types (https://gist.github.com/4432544#file-intrinsics-diff-L30).

Clay's `loadIntrinsicsModule` implementation needs to be updated to correctly populate `__intrinsics__` under LLVM 3.2.

crabtw:
This patch should fix #466.

### DIFF
--- a/compiler/analyzer.cpp
+++ b/compiler/analyzer.cpp
@@ -801,7 +801,7 @@ static MultiPValuePtr analyzeExpr2(ExprPtr expr, EnvPtr env)
             desugarFieldRef(x, safeLookupModule(env));
         if (x->isDottedModuleName)
             return analyzeExpr(x->desugared, env);
-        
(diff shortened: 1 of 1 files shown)


jckarter:
Thanks for doing this @crabtw! Patch looks good to me.

ghost:
Merged, thanks @crabtw .

PR #484, opened 2013-01-22
preliminary support for MIPS32 target

crabtw:
The MIPS O32 ABI implementation is translated from clang's.

### DIFF
--- a/compiler/clay.hpp
+++ b/compiler/clay.hpp
@@ -26,6 +26,7 @@
 #pragma warning(disable: 4146 4244 4267 4355 4146 4800 4996)
 #endif
 
+#include <llvm/ADT/FoldingSet.h>
 #include <llvm/ADT/SmallString.h>
 #include <llvm/ADT/SmallVector.h>
 #include <llvm/ADT/StringMap.h>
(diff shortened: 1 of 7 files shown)


jckarter:
I don't know the MIPS ABI, but the change looks good. Are you able to run clay's lib-clay/externals test suite on a MIPS target?

Longer term I think it'd be a good idea to link libclangCodeGen and reuse Clang's TargetInfo directly instead of reimplementing it case by case. Absorbing clang would also be useful for other things, like invoking the linker without shelling out to a clang process, or building C import functionality directly into the compiler.

crabtw:
Yes, I tested on our MIPS system.
All tests in lib-clay/externals work except x86_float80 and x86_sse.

jckarter:
Cool. One small thing: instead of adding a 'reinitialize' method to the ExternalTargetInfo, it might be cleaner to factor out a CacheableExternalTargetInfo and change the x86 ABIs to subclasses of that, or else simply add the argument offset to the cache key.

crabtw:
Every time we compute foreign function types, the argument offset has to be reset.
However, the overridable methods in `ExternalTarget` don't know if the type is the first one.

I have no idea what is the best way to do this so I insert `reinitialize()`  to `pushReturnType`, `allocReturnValue`, `loadStructRetArgument` methods.
When it tries to compute the return type, it resets the argument offset.

crabtw:
I take some ideas from clang and adds ExternalFunction.
ExternalTarget implements target ABI and generates ExternalFunctions.
And ExternalFunction uses computed results to generate LLVM code.

jckarter:
That's definitely a lot cleaner than the old hacky implementation. Looks great.

PR #486, opened 2013-01-22
set TargetOptions before passing to createTargetMachine

crabtw:
When testing MIPS ABI, I found that LLVM generated FP instructions even though soft-float was set.

It seems that options do not have any effect if we set them after createTargetMachine.


jckarter:
Looks good, thanks.

